PR #72, opened 2014-05-23
A number of fixes to compile with VS2012 and Xcode5

hvellyr:
I would like to propose a number of changes to 

  - fix compiler warnings when building with VS2012 and Xcode5
  - fix a compiler error when using newer versions of boost (tested against 1.53)

In addition I would suggest a small API extension to internal/Step/StepManager.hpp, which allows to access the number of arguments.

Thank you for considering.


paoloambrosio:
Closing in favour of #107 that is rebased and cherry-picked to avoid conflicts with #101 already merged.

PR #73, opened 2014-06-10
use boost option parser for port

Lemutar:
use boost option parser for port, 
so if the user use program options in his test, he will not be forced to use the port parameter


paoloambrosio:
Superseded by #103

Issue #75, opened 2014-12-31
It should be possible to use cucumber-cpp as a submodule


Pooouf:
add_subdirectory(cucumber-cpp cucumber-cpp) should work out of the box

paoloambrosio:
Implemented by #106

PR #76, opened 2014-12-31
Issue #75 : Make it easier to use cucumber-cpp as a submodule

Pooouf:
Support add_subdirectory(cucumber-cpp cucumber-cpp)
Set NOT CUKE_DISABLE_UNIT_TESTS to leave out unit tests


paoloambrosio:
Merged as #106, to re-run the CI build and add cmake cache parameter.

PR #77, opened 2015-01-06
FIX: Invalid submodule URL

APTriTec:
Hello,

Submodule cucumber-tck is referenced with an invalid URL. We are not able to fetch this submodule.

Thanks for your library


APTriTec:
Sorry the issue was due to our firewall.

PR #82, opened 2015-11-02
fix for gtest (October 2015)

ursfassler:


konserw:
I believe that this PR can be closed as it is already merged, probably with one of my PRs

PR #85, opened 2016-01-07
Qt5

konserw:
Adjusted CalcQt example to compile and run with Qt5.
Tested with current stable Qt, version 5.5.1


konserw:
I've added also GTest based steps for CalcQt - I hope it's OK with you.
I think this also fixes issue #83 as i've changed calcqt library to libcalcqt.

konserw:
I've created better PR:
Qt5 support for CalcQt example #98

PR #86, opened 2016-01-10
Setup OSX CI & use externalproject for GMock and GTest & cucmber 1.3.20

konserw:
Please consider compilation of own gmock and gtest using externalproject feature of CMake. I used version gmock 1.7 for this purpose. In my opinion it will be less troublesome. It may also help your efforts to setup windows CI.
Tested only on linux though. 
As you can see now osx travis is up and running. 
Also included some other tempering with travis and script for build and testing.
Cucumber version uplifted to 1.3.20 as 2.x is not recognizing steps.


paoloambrosio:
Thanks for the pull request.

Some commits in this pull request are very good, others need work (e.g. the build runs, fails but it is marked as successful). The major issue I see is that so many unrelated changes make this pull request very difficult to review. The list of changes in the description should be a smell that this PR should be split.

I'll try and comment more later. In the meantime please take a look at this thread on "Cukes Devs" for more information: https://groups.google.com/forum/#!topic/cukes-devs/xRaGM_AEl4U

include/cucumber-cpp/internal/step/StepManager.hpp
```diff
@@ -157,7 +162,7 @@ class StepManager {
 };
 
 
-static std::string toSourceString(const char *filePath, const int line) {
+static inline std::string toSourceString(const char *filePath, const int line) {
```
paoloambrosio: what kind of performance improvement do you expect by inlining this? or was it done for a different reason?

include/cucumber-cpp/internal/step/StepManager.hpp
```diff
@@ -62,6 +62,11 @@ class InvokeArgs {
 
     template<class T> T getInvokeArg(size_type i) const;
     const Table & getTableArg() const;
+
+    size_type argsSize() const {
+      return args.size();
+    }
+
```
paoloambrosio: is this used anywhere?

include/cucumber-cpp/internal/hook/HookRegistrar.hpp
```diff
@@ -35,7 +35,7 @@ class BeforeHook : public Hook {
 
 class AroundStepHook : public Hook {
 public:
-    virtual void invokeHook(Scenario *scenario, CallableStep *step);
+    virtual void invokeHookWithStep(Scenario *scenario, CallableStep *step);
```
paoloambrosio: why this name change? there is only one invokeHook and I don't see the point of renaming it

.travis.yml
```diff
@@ -1,11 +1,27 @@
...
+rvm:
+  - 1.9.3
+  - 2.0
+  - 2.2
+addons:
+  apt:
+    # List of whitelisted in travis packages for ubuntu-precise can be found here:
+    #   https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-precise
+    # List of whitelisted in travis apt-sources:
+    #   https://github.com/travis-ci/apt-source-whitelist/blob/master/ubuntu.json
```
paoloambrosio: don't think comments are needed here: the whitelisting process is quite clear in the Travis documentation

paoloambrosio:
Using GTest/GMock's CMake file was on my list of improvements so I fully support it. It is worth opening an issue (#91) and sending a pull request for that only, but please after we have fixed the build.

.travis.yml
```diff
@@ -1,11 +1,27 @@
...
+os:
+  - linux
+  - osx
 compiler:
   - clang
   - gcc
+rvm:
+  - 1.9.3
+  - 2.0
+  - 2.2
```
paoloambrosio: it does not seem to work (always runs with 2.2.3), probably because it's a c++ project

.travis.yml
```diff
@@ -1,11 +1,27 @@
 language: cpp
-
+sudo: required
+dist: trusty
+os:
+  - linux
+  - osx
```
paoloambrosio: let's create a pull request only for OSX (it's not working)

paoloambrosio:
Compiler warning fixes are very good, can you send a separate PR when we are done with the build maintenance?

.travis.yml
paoloambrosio: After all Ruby deps are there just to run our tests. We shouldn't be concerned with testing multiple Ruby versions. Let's just choose one thst works with rvm (.ruby-version).

konserw: Yes - I noticed that it doesn't work. And agree that it is probably because it's not ruby project for travis. I left it there to think about it later, but then forgot about that...
I agree that we shouldn't be concerned with testing multiple Ruby versions. 
I believe we can just depend on ruby version preinstalled on CI (if it works)

.travis.yml
konserw: OK

konserw:
>Using GTest/GMock's CMake file was on my list of improvements so I fully support it. It is worth opening an issue (#91) and sending a pull request for that only, but please after we have fixed the build.

OK

>Compiler warning fixes are very good, can you send a separate PR when we are done with the build maintenance?

Most of them come from pull request #72 I think. 
 
>Some commits in this pull request are very good, others need work (e.g. the build runs, fails but it is marked as successful). The major issue I see is that so many unrelated changes make this pull request very difficult to review. The list of changes in the description should be a smell that this PR should be split.

I'm sorry for this mess - it's my first attempt to upstream any code. 
Please just disregard this pull request. I'll try to prepare separate cleaner pull requests.

.travis.yml
konserw: Of course. TBH it was copy-paste from google-test .travis.yml

include/cucumber-cpp/internal/hook/HookRegistrar.hpp
konserw: It is from pull request #72

include/cucumber-cpp/internal/step/StepManager.hpp
konserw: from pull request #72

include/cucumber-cpp/internal/step/StepManager.hpp
konserw: from pull request #72

include/cucumber-cpp/internal/step/StepManager.hpp
hvellyr: Hi, sorry to chime in here.  I've used this extension in other code, not pushed or published anywhere (yet).  I needed it to pass it on to another code layer (if I recall correctly).

paoloambrosio: I'd say it belongs to a PR of its own (but it's hard to know if it makes sense without knowing what it should be used for) or it should live in your own branch. It didn't make any sense fixing the compiler warnings or in the OSX build.

gck-ableton: Sure.  I guess I've to rebase my stuff on master once all this and other fixes landed.  I've to review my code (which laid dormant for some time now), and will check what this function was required for.  In the context of the other changes this should be dropped therefore (I guess)

paoloambrosio: Apart from the Boost improvements that should have come from another PR, I think this is the only change that I didn't apply in #107. Once that is merged, it's probably easier to apply this single change on top of master instead of rebasing (that took me quite some time).

Issue #88, opened 2016-01-13
Upgrade to the latest Ruby version and libraries


paoloambrosio:
This should NOT require changing the Travis configuration. It should just work upgrading Ruby dependencies to the latest version and fixing Ruby step definitions.

Step definitions in Cucumber-TCK support an old version of Aruba. Could either fix that or copy it inside the project like the other implementations seem to have.

paoloambrosio:
More info on #87

konserw:
So what are we doing with cucumber-tck? It's quite crucial to decide as I've encountered more problems with it when working on CI for Windows and OSX, see #96 and #97

paoloambrosio:
Good enough for now. Raised #115 to track further work needed.

PR #93, opened 2016-01-18
Upgrade to the latest Ruby version and libraries

konserw:
Upgrade to the latest Ruby version and libraries #88 
This is as far as I could get it. 
Unfortunately I have no experience with ruby or those gems. I'm not even sure if those tests I changed are still properly testing cucumber functionalities.
Maybe I'll try working on own fork of cucumber-tck as with following gems tck itself generates deprected warnings:
-  gem 'cucumber', "=2.0.0"
-  gem 'aruba', "=0.8.0"
-  gem 'rspec', "=3.4.0"


konserw:
I've forked cucumber-tck and fixed warning about #current_dir but that is not sufficient to push aruba further... See own_tck branch for that.

paoloambrosio:
Let's live with `cucumber-tck` till we have something better (still TBD). I've created a `cucumber-cpp` branch in `cucumber-tck` if it helps getting to a stable CI pipeline.

konserw:
as stated also in #97 - I've created PR there: https://github.com/cucumber/cucumber-tck/pull/19

konserw:
i'll change ref here after tck PR is merged

paoloambrosio:
Merged #105 (with 16 less commits)

Issue #91, opened 2016-01-14
Use GTest/GMock CMakeLists instead of FindGTest/FindGMock


paoloambrosio:
It should read like before, but including gtest/gmock's cmake file instead. Users should also be able to configure the path where their version of GTest has been installed, without downloading it (e.g. Ubuntu comes with a package).

Look at changes in #86

konserw:
Debian and ubuntu packages are source only, as google does not recomend
distribution o binary shared libraries. They recommend compilation from
source for each project as some compile-time flags may broke gtest runetime
14 sty 2016 08:26 "Paolo Ambrosio" <notifications@github.com> napisał(a):

paoloambrosio:
Yes indeed. For that reason, instead of forcingly downloading 1.7.0 in CMake we should give the OPTION for the user to specify where GTest/GMock sources are present, instead of assuming that the user does not know what he/she is doing.

konserw:
Point taken.

2016-01-14 13:57 GMT+01:00 Paolo Ambrosio <notifications@github.com>:

konserw:
I'm working on it in branch new_gtest feel free to comment it :)

Issue #89, opened 2016-01-13
Upgrade C++ dependencies


paoloambrosio:
- Latest library version should work (or a reasonably recent depending on what can be supported on CI servers)
- Earliest supported must be those in the latest Ubuntu stable (Boost 1.54, Google Mock/Test 1.6.0), latest Debian stable (Boost 1.55, Google Mock/Test 1.7.0), ...

Examples are a special case and in general they might just support the latest. Upgrade to Qt 5 is not mandatory at this stage.

Different versions might be tested on different platforms (we shouldn't be too picky at this stage).

konserw:
GMock 1.6.0 fails - see https://travis-ci.org/konserw/cucumber-cpp/jobs/104175912 - i'm not sure how to fix it?

konserw:
GMock 1.6 issue fixed with commit 47c8be604e43bea729c6b7c35d7bcbdf55966fa5

kai-unger:
I adapted the boost driver to support the latest boost version 1.60 in my [fork of cucumber-cpp](https://github.com/kai-unger/cucumber-cpp).

I tested it with ctest and running BoostCalculatorSteps with cucumber 2.0.0 on OS X 10.11.3 (system compiler clang) with boost versions 1.54, 1.58, 1.59 and 1.60. #101

PR #94, opened 2016-01-23
Use GTest/GMock CMakeLists instead of FindGTest/FindGMock

konserw:
Use GTest/GMock CMakeLists instead of FindGTest/FindGMock #91 
This pull request contain commit from PR #82


paoloambrosio:
Can you put all the logic to find or download GTest/GMock into a separate module file (in cmake/modules like the previous one for GMock)? Also in that file please add a preamble like in https://github.com/Kitware/CMake/blob/master/Modules/FindGTest.cmake (this should include also GTEST_ADD_TESTS that you put into tests/CMakeLists.txt)

Google's project on GitHub is called googletest so I would focus on that instead of GoogleMock. It is also standard to use GTEST_ROOT to tell CMake where to find libraries. Suggested variable naming: CUKE_GMOCK_DIR -> GTEST_ROOT, CUKE_GMOCK_VER -> GTEST_VER? (the module should not contain any reference to Cucumber-Cpp and it should be generic)

I would also like it to export the same variables as the original GMock/GTest (GTEST_FOUND, GTEST_MAIN_LIBRARIES, GTEST_LIBRARIES, GMOCK_FOUND, ...) so that it can be used in the same way (e.g. 
if(GTEST_FOUND) instead of if(NOT CUKE_DISABLE_GTEST)).

README.md
```diff
@@ -18,12 +18,12 @@ It relies on a few libraries:
 * [Boost](http://www.boost.org/) 1.40 or later.
   Required libraries: *thread*, *system*, *regex*, and *date_time*.
   Optional library for Boost Test driver: *test*.
-* [GTest](http://code.google.com/p/googletest/) 1.4 or later.
-  Optional for the GTest driver.
+* [GTest](http://code.google.com/p/googletest/) 1.6 or later.
```
paoloambrosio: Aren't we supporting GTest 1.4 anymore?

include/cucumber-cpp/internal/step/StepManager.hpp
```diff
@@ -43,7 +43,7 @@ class MatchResult {
     void addMatch(SingleStepMatch match);
 
     operator void *();
-    operator bool();
+    operator bool() const;
```
paoloambrosio: Should this be part of this PR?

src/StepManager.cpp
```diff
@@ -44,7 +44,7 @@ MatchResult::operator void *() {
     return (void *) resultSet.size();
 }
 
-MatchResult::operator bool() {
+MatchResult::operator bool() const {
```
paoloambrosio: Should this be part of this PR?

tests/CMakeLists.txt
```diff
@@ -1,3 +1,37 @@
+# Thanks to Daniel Blezek <blezek@gmail.com> for the GTEST_ADD_TESTS code
```
paoloambrosio: You can reference where you copied it from - CMake's FindGTest - instead of the author of this function that is already cited there

travis.sh
```diff
@@ -1,8 +1,13 @@
 #!/bin/sh
 set -e #break script on non-zero exitcode from any command
 gem install bundler
+
 cmake -E make_directory build
-cmake -E chdir build cmake -DCUKE_ENABLE_EXAMPLES=on -DGTEST_INCLUDE_DIR=/usr/src/gmock/gtest/include/  ..
+if [ -z ${GMOCK_PATH+x} ]; then 
+    cmake -E chdir build cmake -DCUKE_ENABLE_EXAMPLES=on -DCUKE_GMOCK_VER=${GMOCK_VER}  .. ;
```
paoloambrosio: No need for semicolons before a new line

konserw: ok, but doesn't hurt either AFAIK

tests/CMakeLists.txt
konserw: My bad

include/cucumber-cpp/internal/step/StepManager.hpp
konserw: not sure TBH

src/StepManager.cpp
konserw: not sure TBH

README.md
konserw: You said in #89 :

> Earliest supported must be those in the latest Ubuntu stable (Boost 1.54, Google Mock/Test 1.6.0), latest Debian stable (Boost 1.55, Google Mock/Test 1.7.0), ...

I've understood that 1.6.0 should be oldest supported. 
Haven't tested with older version, also haven't added to travis. But i don't see any reason the same code wouldn't work with older versions - just like stated in README:

>   It might work with earlier versions of the libraries, but it was not
 tested with them.

paoloambrosio: As you said, it should work with older releases (1.4) but really 1.6 has been released in 2011 so let's go with that! Perhaps we can add "It might work, but it is not guaranteed, to work with earlier versions of those libraries", or something on those lines?

konserw: It's already there:


    It might work with earlier versions of the libraries, but it was not
    tested with them.

paoloambrosio: Cool. You can ignore my comment then :-)

konserw:
Something like this ?

konserw:
I believe now it's ready for next review.

konserw:
I've added some fixes for building with Visual Cpp and option for more verbose output. There is now spearate GMOCK_FOUND and GTEST_FOUND

.travis.yml
```diff
@@ -14,16 +18,9 @@ addons:
...
- 
-before_script:
-  - "cd /usr/src/gmock && sudo cmake . && sudo cmake --build . && sudo mv libg* /usr/local/lib/ && sudo mv gtest/libg* /usr/local/lib/ ; cd -"
 
 script: ./travis.sh 
 
-branches:
-  only:
-    - master
-
```
paoloambrosio: Why has it been removed?

paoloambrosio:
Sorry for the delay in reviewing this PR. Apart from my last comment, it looks ready to be merged!

.travis.yml
konserw: Because it stopped my branch from building on travis. It's irrelevant to this repository as it hasn't got any active branches.

konserw:
Good, as it's needed for appveyor PR I'm preparing ;)

.travis.yml
paoloambrosio: Not sure what this adds but I guess it's fine.

I'll try and find a way of sending emails to the group by default only for what builds from cucumber/cucumber-cpp repo. Since it wasn't a PR, your appveyor branch in konserw/cucumber-cpp shouldn't have sent notifications there when failing.

I'll merge it ASAP.

paoloambrosio:
The end result was very good

This PR closes #91

.travis.yml
brasmusson: Yes, it is true that this limited the branches build on Travis to master only. To have other branches build (before creating a PR for instance), my method would have been that on that branch add that branch to the list of branches to build, and then remove it again before creating the PR.

PR #98, opened 2016-02-13
Qt5 support for CalcQt example

konserw:
Now CalcQt example should work on both Qt4 and Qt5.
I think this also fixes issue #83 as i've changed calcqt library to libcalcqt.


konserw:
merge

examples/CalcQt/CMakeLists.txt
```diff
@@ -1,21 +1,49 @@
 project(CalcQt)
 
-find_package(Qt4 COMPONENTS QtCore QtGui QtTest)
+set(CALCQT_HEADERS src/CalculatorWidget.h)
+set(CALCQT_SOURCES src/CalcQt.cpp src/CalculatorWidget.cpp)
+set(LIBCALCQT_BUILD OFF)
```
paoloambrosio: Is `LIBCALCQT_BUILD` really needed? If it's for the `if(LIBCALCQT_BUILD)` later on, can't you use the `QT_LIBRARIES` variable?

paoloambrosio:
Nice PR! I trust you have tested it with QT5, and I have with QT4. Apart from the comment it should be ready to merge.

examples/CalcQt/CMakeLists.txt
konserw: Good idea ;) Let's check that.

konserw:
Ok, now it should be ready for merging if travis tests passes.

Issue #80, opened 2015-06-02
Error when no testing framework is detected


paoloambrosio:
When the user does not include a testing framework before the Cucumber-CPP include, it falls back to the FakeDriver. This should be renamed to GenericDriver and enabled only if explicitly requested by defining a macro, otherwise it should error,

The implemententation should also be extracted from the include or a step in a second file would error with:
```
cucumber-cpp/internal/drivers/FakeDriver.hpp:17: multiple definitions of
«cucumber::internal::FakeStep::invokeStepBody() »
```

paoloambrosio:
Implemented in #99

PR #99, opened 2016-02-13
Generic driver

konserw:
My proposition for issue:
 Error when no testing framework is detected #80


konserw:
> In file included from /home/konserw/git/cucumber-cpp/include/cucumcucumberber-cpp/internal/drivers/DriverSelector.hpp:9:0,
>                  from /home/konserw/git/cucumber-cpp/include/cucumber-cpp/defs.hpp:6,
>                  from /home/konserw/git/cucumber-cpp/features/step_definitions/RpnCalculatorSteps.cpp:2,
>                  from /home/konserw/git/cucumber-cpp/build/tmp/test_features/step_definitions/cpp_steps.cpp:15:
> /home/konserw/git/cucumber-cpp/include/cucumber-cpp/internal/drivers/GenericDriver.hpp:9:7: error: redefinition of ‘class cucumber::internal::FakeStep’
>  class FakeStep : public BasicStep {
>        ^
> In file included from /home/konserw/git/cucumber-cpp/include/cucumber-cpp/internal/drivers/DriverSelector.hpp:9:0,
>                  from /home/konserw/git/cucumber-cpp/include/cucumber-cpp/defs.hpp:6,
>                  from /home/konserw/git/cucumber-cpp/build/tmp/test_features/step_definitions/cpp_steps.cpp:2:
> /home/konserw/git/cucumber-cpp/include/cucumber-cpp/internal/drivers/GenericDriver.hpp:9:7: error: previous definition of ‘class cucumber::internal::FakeStep’
>  class FakeStep : public BasicStep {
>        ^
> 

If anyone can see why this happens please help me ;)

konserw:
Ok so now my proposition works as expected.
Any comments?

include/cucumber-cpp/internal/drivers/DriverSelector.hpp
```diff
@@ -1,9 +1,13 @@
...
+    #include "BoostDriver.hpp"
 #elif defined(CPPSPEC_H_)
-#include "CppSpecDriver.hpp"
+    #include "CppSpecDriver.hpp"
 #else // No test framework
-#include "FakeDriver.hpp"
+    #ifdef CUKE_ENABLE_GENERICDRIVER
+        #include "GenericDriver.hpp"
+    #else
+        #error Please #inlude testing framework before cucumber-cpp or #define CUKE_ENABLE_GENERICDRIVER explicitly
```
paoloambrosio: Typo `#inlude` => `#include`

include/cucumber-cpp/internal/drivers/GenericDriver.hpp
```diff
@@ -0,0 +1,19 @@
+#ifndef CUKE_GENERICDRIVER_HPP_
+#define CUKE_GENERICDRIVER_HPP_
+
+#include "../step/StepManager.hpp"
+
+namespace cucumber {
+namespace internal {
+
+class FakeStep : public BasicStep {
```
paoloambrosio: Forgot to rename it to `GenericStep`?

paoloambrosio:
PR looks OK (see comments) but I have to say that I don't like much the macro `CUKE_ENABLE_GENERICDRIVER`. What do you think of triggering behaviour based on a different include? It shouldn't require many changes.

**Current**

With testing framework:
```
#include <CppSpec/CppSpec.h>
#include <cucumber-cpp/defs.hpp>
```
Without testing framework:
```
#include <cucumber-cpp/defs.hpp>
```

**Proposed**

With testing framework (it errors if it can't find it):
```
#include <CppSpec/CppSpec.h>
#include <cucumber-cpp/autodetect.hpp>
```
Without testing framework (using the generic driver):
```
#include <cucumber-cpp/generic.hpp>
```

Under the hood, `autodetect.hpp` would include `DriverSelector.hpp` and that should error if it can't find it, instead `generic.hpp` would include `GenericDriver.hpp`.

paoloambrosio:
...or perhaps instead of introducing `#include <cucumber-cpp/generic.hpp>` we could keep using `#include <cucumber-cpp/defs.hpp>` if someone does not want the safety net of `autodetect.hpp`?

**Revised proposal**

With testing framework (it errors if it can't find it):
```
#include <CppSpec/CppSpec.h>
#include <cucumber-cpp/autodetect.hpp>
```
Without testing framework (using the generic driver as fallback if not found - backward compatible):
```
#include <cucumber-cpp/defs.hpp>
```

include/cucumber-cpp/internal/drivers/GenericDriver.hpp
konserw: You've mentioned only class name in #80 - wasn't sure if this should also be changed

paoloambrosio:
Added a test for the Generic Driver and removed duplication in paoloambrosio/cucumber-cpp@22c427a. Please include it in this PR.
Also had a change of heart and thought the autoswitching between generic and not was too dangerous, so I've marked it as deprecated in paoloambrosio/cucumber-cpp@cae0f25. For this I'll let you decide what you prefer.

konserw:
I think that both your commits are good ;) I'm testing them right now and I'll push them in a second :)

paoloambrosio:
In cases like 059878d + bd7871a, feel free to remove commits and use "push -f" to revert your own branch/PR so that the history reads better.

Is this ready to merge?

konserw:
I've rebased it once again. I believe now it is ready ;)

PR #100, opened 2016-02-14
Add a getting started section to README.md

MartinDelille:


konserw:
nice! :+1:

README.md
```diff
@@ -63,3 +63,23 @@ start build\examples\Calc\BoostCalculatorSteps.exe
...
+```
+host: localhost
+port: 3902
+```
+
+Create your first feature (an example is available [here](examples/Calc/features/addition.feature)).
+
+Then create your step definition runner (an example is available [here](examples/Calc/features/step_definitions/BoostCalculatorSteps.cpp)). In order to compile the step definition runner, make sure to add [cucumber include directory](includes) to the include path and link with *libcucumber-cpp.a* and additional testing libraries (boost unit test).
+
+Run the step definition runner in the background (adding an ampersand), then `cucumber`. Your feature should be executed (and the step definition runner exit).
```
paoloambrosio: I would rephrase it in a more generic way (not Linux/OSX specific) like...

Run the step definition runner in the background and then cucumber, like in the Calc example in the previous section. The step definition runner should exit after the feature is run and cucumber exits.

paoloambrosio:
Very nice indeed!

PR #97, opened 2016-02-12
Appveyor - windows CI

konserw:
Build cucumber on appveyor - windows CI platform.
Only thing that's not working is features target. I'm not sure how to fix this. Probably it has somethik do to with naming of targets (see test target)

This is rebased, shortened version of my appveyor_old branch - if you want to know how i came up with all this strange things feel free to examine over 50 commits there: https://github.com/konserw/cucumber-cpp/tree/appveyor_old

PR #102, opened 2016-02-29
Fix OSX build

paoloambrosio:
Build fails on OSX because make does not understand that the step definition sources have changed between one scenario and the next, even when the entire directory is removed.

Couldn't find a cleaner way to fix it than to remove the object file before recompiling it.

Can't test it on Windows at the moment.


konserw:
I've also done that in #96 but your code is better, so i'll merge letter.

> Can't test it on Windows at the moment.

You should consider merging #97 then ;)

konserw:
Commit 844b720 broken appveyor build - not sure why

PR #104, opened 2016-03-10
Working OSX CI on Travis

konserw:
Now it's working, see: https://travis-ci.org/konserw/cucumber-cpp/jobs/115146291 Unfortunately only with GTest 1.7. 
Shall we build also with gcc on osx?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,6 @@ Gemfile.lock
 
 # Build artifacts
 build/
+
+#vim swap files
+*.swp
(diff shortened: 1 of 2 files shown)


paoloambrosio:
Yes, you can leave GCC out since on OSX it uses clang under the hood :-)

Can you squash your commits? 10 commits for 15 lines of useful code is a bit excessive! Please when you send PRs don't send rubbish commits that should not be there.

After that, I'll be **very happy** to merge it!

konserw:
Still better then 37 commits in #96 ;)
After squash shall i force push here, or create separate PR?

paoloambrosio:
Please force push here :)

konserw:
Ok, now it should be ok :)

PR #107, opened 2016-03-13
A number of fixes to compile with VS2012 and Xcode5

paoloambrosio:
Cherry-pick of #72 from @hvellyr that now has too many conflicts


paoloambrosio:
I don't feel hvellyr/cucumber-cpp@0e76db6 is the right fix for the compiler warning. I'll try to refactor my messy hooks code to fix it.

paoloambrosio:
84ab6fa **should** fix the method hiding warning (that I can't turn on with gcc!). Unfortunately the hooks implementation is so bad that there is no simple and clean fix.

PR #108, opened 2016-03-14
appveyor windows CI

konserw:
New fixed and rebased version of windows CI

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,5 @@
 version: "{branch}-ci-{build}"
+image: Visual Studio 2015
 os:
 - Visual Studio 2013
 
@@ -13,15 +14,15 @@ on_finish:
 #- ps: $blockRdp = $true; iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
 
 environment:
+  ruby_version: "200"
   matrix:
-    - BOOST_ROOT: C:\Libraries\boost
-      BOOST_INCLUDEDIR: C:\Libraries\boost\boost
-      BOOST_LIBRARYDIR: C:\Libraries\boost\lib32-msvc-12.0
-      ruby_version: "193"
-    - BOOST_ROOT: C:\Libraries\boost_1_59_0
-      BOOST_INCLUDEDIR: C:\Libraries\boost_1_59_0\boost
-      BOOST_LIBRARYDIR: C:\Libraries\boost_1_59_0\lib32-msvc-12.0
-      ruby_version: "200"
+    - BOOST_ROOT: 'C:\Libraries\boost'
+      BOOST_INCLUDEDIR: 'C:\Libraries\boost\boost'
+      BOOST_LIBRARYDIR: 'C:\Libraries\boost\lib32-msvc-12.0'
+    - BOOST_ROOT: 'C:\Libraries\boost_1_58_0'
+      BOOST_INCLUDEDIR: 'C:\Libraries\boost_1_58_0\boost'
+      BOOST_LIBRARYDIR: 'C:\Libraries\boost_1_58_0\lib32-msvc-12.0'
+      
 
 install:
 - git submodule init
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,53 @@
+version: "{branch}-ci-{build}"
+os:
+- Visual Studio 2013
+
+platform:
+- Win32
+
+configuration:
+- Debug 
+#- Release
+
+on_finish:
+#- ps: $blockRdp = $true; iex ((new-object net.webclient).DownloadString('https://raw.githubusercontent.com/appveyor/ci/master/scripts/enable-rdp.ps1'))
+
+environment:
+  matrix:
+    - BOOST_ROOT: C:\Libraries\boost
+      BOOST_INCLUDEDIR: C:\Libraries\boost\boost
+      BOOST_LIBRARYDIR: C:\Libraries\boost\lib32-msvc-12.0
+      ruby_version: "193"
+    - BOOST_ROOT: C:\Libraries\boost_1_59_0
+      BOOST_INCLUDEDIR: C:\Libraries\boost_1_59_0\boost
+      BOOST_LIBRARYDIR: C:\Libraries\boost_1_59_0\lib32-msvc-12.0
+      ruby_version: "200"
+
+install:
+- git submodule init
+- git submodule update
+- SET PATH=C:\Ruby%ruby_version%\bin;%PATH% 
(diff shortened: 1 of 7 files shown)


konserw:
It was broken because Windoze doesn't recognize any signal except of "KILL" [see]() Probably this is M$ "feature" ;/

konserw:
Ok, I've cleaned up this PR, both CI are green, now only thing left is cucumber-tck changes.

paoloambrosio:
Tried it on Windows 10 + VS 2015. I get "C:\Libraries\boost_1_60_0\boost/config/auto_link.hpp(365): fatal error C1189: #error:  "Mixing a dll boost library with a static runtime is a really bad idea..."

fixCucumber.ps1
```diff
@@ -0,0 +1,9 @@
+If ($env:ruby_version -match "200") {
+    $path = "C:\Ruby200\lib\ruby\gems\2.0.0\gems\cucumber-2.0.0\lib\cucumber\formatter\ansicolor.rb"
+} Else {
+    $path = "C:\Ruby193\lib\ruby\gems\1.9.1\gems\cucumber-2.0.0\lib\cucumber\formatter\ansicolor.rb"
+}
+
+(Get-Content $path) -replace '\ST\DERR.puts %', '#nope' | Set-Content $path 
+(Get-Content $path) -replace '    Cucumber::', '#nope' | Set-Content $path
+(Get-Content $path) -replace '^Cucumber::', '#nope' | Set-Content $path
```
paoloambrosio: This seems to fix a problem in Cucumber 2.0.0... is it a documented problem? Can you post a link? Does it happen with Ruby 2.2 as well?

appveyor.yml
```diff
@@ -0,0 +1,53 @@
...
+environment:
+  matrix:
+    - BOOST_ROOT: 'C:\Libraries\boost'
+      BOOST_INCLUDEDIR: 'C:\Libraries\boost\boost'
+      BOOST_LIBRARYDIR: 'C:\Libraries\boost\lib32-msvc-12.0'
+      ruby_version: "193"
+    - BOOST_ROOT: 'C:\Libraries\boost_1_58_0'
+      BOOST_INCLUDEDIR: 'C:\Libraries\boost_1_58_0\boost'
+      BOOST_LIBRARYDIR: 'C:\Libraries\boost_1_58_0\lib32-msvc-12.0'
+      ruby_version: "200"
```
paoloambrosio: I think we don't need Ruby in the dependency matrix. We can use the latest.

appveyor.yml
```diff
@@ -0,0 +1,53 @@
+version: "{branch}-ci-{build}"
+os:
+- Visual Studio 2013
```
paoloambrosio: Is there a reason behind the choice of VS 2013? Can we have multiple OSs?

paoloambrosio:
I also get
```
CMake Warning at CMakeLists.txt:197 (message):
  Could not find Cucumber: skipping end-to-end tests
```
if I don't install Cucumber-Ruby before running CMake (like you do in your script).

fixCucumber.ps1
konserw: It's more of a problem of how appveyor is implemented, i.e. it doesn't need any additional library to get colored output - just like on *nix - and cucumber thinks it is not possible on windows. See:
https://github.com/appveyor/ci/issues/373
http://help.appveyor.com/discussions/suggestions/197-support-ansi-color-codes
It's not dependent on ruby version.

appveyor.yml
konserw: I've just left it from your's appveyor branch ;)
I guess we could check several versions

appveyor.yml
konserw: Ok, I think ruby 2.0.0 is latest that works with cucumber on windows. Not sure why ;)

konserw:
About cuvumber-ruby - have you done bundle install ?

appveyor.yml
konserw: I've found it: 



>  I'm afraid Cucumber on Windows/Ruby requires ruby 2.0.0 (x86). 
>  The reason is that the gherkin gem doesn't yet ship with compiled binaries for more recent versions or Ruby, and not for x64.
> 
>  We're working on a Gherkin3 which will address this issue. 
>  See this post for background. I cannot give you an ETA, but it's at least a few months away.

[source](http://stackoverflow.com/questions/29507570/cannot-load-such-file-2-2-gherkin-lexer-en-how-to-fix)

appveyor.yml
konserw: Ok - we need VS2013 because we have precompiled boost libs only for that version (C:\libraries\boost\lib-win32-vs12 (something like that))

paoloambrosio: Pretty sure every OS comes with its version of Boost (they are different VMs)

konserw: I'm pretty sure I've checked that.

appveyor.yml
brasmusson: Cucumber-ruby v2.1.0 (and later) use Gherkin3, so to use a newer Ruby version on Windows, you need to update to a newer Cucumber-ruby version.

appveyor.yml
paoloambrosio: It must be another Appveyor bug. Boost 1.59 has the right MSVC14 version, the other do not.

konserw: curious. I'm afraid previously boost 1.59 caused me some problems - I'll look into this

konserw: > It must be another Appveyor bug. Boost 1.59 has the right MSVC14 version, the other do not.
Or a feature :) (I'm a QA ;) )
I'm now trying msvc14+boost1.59 in addition to old config oiptions

konserw:
Ok, now I believe it can be merged.

konserw:
Abot that:

> Tried it on Windows 10 + VS 2015. I get "C:\Libraries\boost_1_60_0\boost/config/auto_link.hpp(365): fatal error C1189: #error: "Mixing a dll boost library with a static runtime is a really bad idea..."

I've also done some tests and I don't know how to enable boost test on windows properly - this is why i've added -DCUKE_DISABLE_BOOST_TEST in my appveyor config.

All in all it seems to me that boost test is really problematic to use...

paoloambrosio:
Spent a few (horrible) days on Windows to fix the build. Created and merged #113 and then the AppVeyor build #114 was simple. Sorry to reject this PR.

Notes:
 - The E2E suite just needed the full path for Cucumber, without changing `run_simple`
 - Didn't add any hack for build colours, as they are not important to justify a patch
 - Not sure the build matrix was working since changing the CMake generator should not change the VS used or the architecture (x64 needs Boost lib64, so my guess is that x86 was always used)

PR #110, opened 2016-03-18
QtTest driver

konserw:
It's QtTest based test driver for cucumber-cpp.
I'm not sure if unit test for that is OK.
Looking forward for any comments ;)


konserw:
I've rebased to clean up commit history - now I believe it's ready for merging.
I've switched travis build to qt5 as qttestdriver I've created is qt5-only (pretty sure it is easy to port to older qt also) - should we add both qt4 and 5 to travis build matrix ?

tests/integration/drivers/QtTestDriverTest.cpp
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+private:
+    void stepInvocationInitsQtTest() {
+        QtTestSteps framework;
+        QCOMPARE(framework.isInitialized, false);
+        QCOMPARE(framework.isCleanedUp, false);
+        QTest::qExec(&framework);
+        QCOMPARE(framework.isInitialized, true);
+        QCOMPARE(framework.isCleanedUp, true);
```
paoloambrosio: This is not testing the driver

konserw: I've used boost test as example - how this is different?

```
class BoostDriverTest : public DriverTest {
public:
    virtual void runAllTests() {
        stepInvocationInitsBoostTest();
        DriverTest::runAllTests();
    }

private:
    void stepInvocationInitsBoostTest() {
        std::cout << "= Init =" << std::endl;
        using namespace boost::unit_test;
        BoostStepDouble step;
        expectFalse("Framework is not initialized before the first test", framework::is_initialized());
        step.invokeStepBody();
        expectTrue("Framework is initialized after the first test", framework::is_initialized());
    }
};

int main(int argc, char **argv) {
    BoostDriverTest test;
    return test.run();
}
```

paoloambrosio: The tests for the Boost driver exercise a test double that inherits from the class under test (instead of exercising the framework directly). Assertions instead check properties of the test framework to make sure that the .invokeStepBody() method did something on the system.

`class BoostStepDouble : public BoostStep {` VS `class QtTestSteps : public QObject {`
`step.invokeStepBody();` VS `QTest::qExec(&framework);`

src/drivers/QtTestDriver.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+    QTemporaryFile file;
+    QString fileName;
+    if (file.open()) {
+        fileName = file.fileName();
+    }
+    file.close();
+
+    QtTestObject test(this);
+    int returnValue = QTest::qExec(&test, QStringList() << "test" << "-o" << fileName);
```
paoloambrosio: I have to say I'm not thrilled about writing a file for each scenario step. Isn't there any other way? Perhaps diving deeper into the implementation? I see QTest::qExec doing plenty of things that are not needed by a step run.

.travis.yml
```diff
@@ -19,6 +19,8 @@ matrix:
 
 addons:
   apt:
+    sources:
+      - ubuntu-toolchain-r-test
```
paoloambrosio: It might be a stupid question, but why is this needed?

paoloambrosio:
Apart from the comments, nice PR!

paoloambrosio:
...and don't forget to add something in the library dependencies of the README file!

.travis.yml
konserw: For the qt5 package I think ;)

src/drivers/QtTestDriver.cpp
konserw: I've found nice way to get Qt's output piped directly on POSIX compilant systems, but not on windows ;/ Please see 80408f0

konserw:
I've updated readme, rebased it and partly fixed file writting problem. 
Could you give me some advice about how to produce good driver test?

src/drivers/QtTestDriver.cpp
konserw: Also I do not think diving any deeper into implementation is possible as we would need to use Qt's private API which can change between minior versions

konserw:
Any idea on why would this test pass on gcc and fail on clang?

.travis.yml
konserw: I've removed it. That repo was needed to get qt5 packages on ubuntu 14.04 at some point in history I think, not anymore.

konserw:
Stupid mistake - those bool variables were uninitialized, that defaults to false (0) on gcc, but not on clang

konserw:
Anyone any comments?

konserw:
If no comments merge please :)

konserw:
Just rebased against current master

konserw:
Rebased and all green - ready to merge ;)

konserw:
One last commit - for compatibility with Qt >= 5.7 c++11 has to be enabled. I wasn't sure if new PR would be better for this...

paoloambrosio:
I have to say I don't like the Posix implementation at all:
 - don't like temprarily redirecting the entire process's stdout
 - it has been copied from stack overflow, that is at the very top of my list of things no developer should do

You are right: there is no way out of the horribly closed implementation that Qt has done (shame on them!). On a Posix system we could create a new process running the same code base (fork) and read from its output, but that is not supported by Windows. At this point the lesser of all evils is to use the temporary file solution for all implementations :-(

src/drivers/QtTestDriver.cpp
```diff
@@ -0,0 +1,98 @@
...
+        read(out_pipe[0], buffer, maxLength); /* read from pipe into buffer */
+        return InvokeResult::failure(buffer);
+    }
+#else
+    QTemporaryFile file;
+    QString fileName;
+    if (file.open()) {
+        fileName = file.fileName();
+    }
+    file.close();
```
paoloambrosio: If the file cannot be opened, the filename will be null and nothing would work. Perhaps we should stop here and return a failure.

src/drivers/QtTestDriver.cpp
```diff
@@ -0,0 +1,98 @@
...
+
+    int returnValue = QTest::qExec(&test, QStringList() << "test" << "-o" << fileName);
+    if(returnValue == 0)
+        return InvokeResult::success();
+    else
+    {
+        file.open();
```
paoloambrosio: If we close the file at the end, we don't need to re-open it. Also please mind the coding conventions (spaces, braces, ...):
```
if (...) {
    ...
} else {
    ...
}

File: src/drivers/QtTestDriver.cpp [hunk 6]
```diff
@@ -0,0 +1,98 @@
...
+        QTextStream ts(&file);
+        return InvokeResult::failure(ts.readAll().toLatin1());
```
paoloambrosio: Why `toLatin1`? I'm no Qt expert but `toLocal8Bit` looks more like the OS's character set... am I wrong?

include/cucumber-cpp/internal/drivers/QtTestDriver.hpp
```diff
@@ -0,0 +1,28 @@
...
+#define CUKE_QTTESTDRIVER_HPP_
+
+#include "../step/StepManager.hpp"
+
+namespace cucumber {
+namespace internal {
+
+class QtTestStep : public BasicStep{
+public:
+    QtTestStep(): IsInitialized(false), IsCleanedUp(false), TestRun(false) {}
```
paoloambrosio: Is there a reason why here and in several other places you are using UpperCamelCase instead of lowerCamelCase?

paoloambrosio:
I'm not sure but I believe that there could be an alternative to writing a real file, that is to write the output to a named pipe (available on both Windows and Posix systems, but with different APIs). Happy to merge this with file based I/O as a multi-platform pipe implementation is not straightforward.

konserw:
Please take a look at new implementation using pipes for both POSIX and windows.
It doesn't address both your considerations as this is also mostly copied from SO and it's redirecting entire stdout, but for sure it is better than previous implementation. If you still don't like it I'll prepare temporary-file based version.
I've also rebased against current master and enabled Qt to be found by CMake on windows. Unfortunately travis OSX image does not contain Qt, and building it from source would be quite a burden I think, see: https://github.com/travis-ci/travis-ci/issues/5278
By the way I moved building commands from appveyor.yml to appveyor.bat

konserw:
I've also experimented with moving qttest::qwxec to another thread but then first step from calcqt example allways fail (widget does not render, so label is empty), you can check here: https://github.com/konserw/cucumber-cpp/tree/qttestthread
console apps seem to work, the problem is CalculatorWidget that should be created in main thread (also known in qt as GUI thread... )

marco-piccolino:
@konserw for pre-builded Qt on Travis CI you can take a look here: https://github.com/benlau/qtci

konserw:
I'm closing this PR as I've splitted it into 2 new:
#140 for CI and Cmake base for just CalcQt example
#142 for driver itself

PR #111, opened 2016-03-20
Fixed documentation

maxmeyer:
As discussed the upgrade and the outcome of our discussion at #cukenhagen.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,12 +46,23 @@ to run the functional test suite.
 Building Cucumber-Cpp with tests and samples:
 
 ```
+# Download test suite
 git submodule init
 git submodule update
+
+# Create build directory
 cmake -E make_directory build
+
+# Generate Makefiles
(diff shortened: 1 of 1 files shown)


konserw:
A little misleading title as it has nothing to do with aruba ;)

paoloambrosio:
@konserw: correct, the two commits have nothing to do with the title and might be moved to a separate PR

@maxmeyer is the maintainer of Aruba and has offered to help upgrading from 0.8 to 0.14!

Late in the evening, after a day of hacking, we were too tired to write the first step of something that could be committed, but at least now I know one of the problems and I'm working on it ;-)

konserw:
Ok - good to hear that ;)

maxmeyer:
@konserw Fixed the title. We rolled back some changes. So nothing of the upgrade was left.

PR #116, opened 2016-07-18
Issue79

konserw:
Fix for issue #79 - Suggested step definition when step sentence contains double quote

PR #120, opened 2016-09-07
Download gmock/gtest from github

konserw:
Download gmock/gtest from github as link used currently expired

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ compiler:
   - clang
   - gcc
 env:
+  - GMOCK_VER=1.8.0 
   - GMOCK_VER=1.7.0 
   - GMOCK_PATH=/usr/src/gmock #1.6.0 from ubuntu trusty repo
 matrix:
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,11 +84,13 @@ if(NOT CUKE_DISABLE_GTEST)
     set(GMOCK_ROOT "${CMAKE_CURRENT_BINARY_DIR}/gmock")
     find_package(GMock REQUIRED)
 
-    if(GMOCK_FOUND)
+    if(GTEST_FOUND)
         set(CUKE_GTEST_LIBRARIES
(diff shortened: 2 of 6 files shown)


konserw:
Can you see why would most features fail because of tempering with gtest/gmock ?

konserw:
ok I think I've got it - for some reason if GMOCK_ROOT is not a subdirectory it messes e2e tests

konserw:
I've added support for gmock 1.8.0 to FindGmock but we have issue #121

srbaker:
+1 on this. Would love to see this merged soon, as master won't even build currently. :(

paoloambrosio:
Nice one. Haven't gone through all lines changed, but I trust the CI pipeline.

I'd squash the first 4 commits into "change gmock/gtest source to github" and keep "support for gtest/gmock >= 1.8.0".

A couple of questions:
 - What about compiling with 1.8.0 in Travis / AppVeyor?
 - Why is 1.7.0 still the default?

paoloambrosio:
@srbaker it does compile if you specify `GMOCK_PATH` (see https://travis-ci.org/cucumber/cucumber-cpp/jobs/158838935). The CI pipeline doesn't currently run on master because it also tries to compile with the sources downloaded from GitHub.

konserw:
I had some problems with gmock 1.8, see issue #121 - i'll need to verify what happened there :)

paoloambrosio:
If it doesn't automaticaly work with 1.8.0 we should split the two PRs and remove last commit from this one. In the one for 1.8.0 please add that to the CI build. After that and squashing I think it can be merged to close #122 (and finally merge #116).

konserw:
@paoloambrosio - I've fixed building with gmock 1.8 and squashed into 2 commits. 
 I've also added third commit that adds gmock 1.8.0 to travis CI

konserw:
all green :)

paoloambrosio:
Nice one! Green build again :+1:

PR #124, opened 2016-10-20
Fixing non valid preprocessor warning on msvc

aallrd:
The `#warning` keyword is not understood by the msvc preprocessor, throwing the below fatal error:
`fatal error C1021: invalid preprocessor command 'warning'` 
This fix provides msvc portability for the preprocessor warning message.

### DIFF
--- a/include/cucumber-cpp/defs.hpp
+++ b/include/cucumber-cpp/defs.hpp
@@ -1,4 +1,8 @@
-#warning Use of defs.hpp is deprecated, please use either autodetect.hpp or generic.hpp
+#ifdef __GNUC__
+    #warning "Use of defs.hpp is deprecated, please use either autodetect.hpp or generic.hpp"
+#else
+    #pragma message( "Use of defs.hpp is deprecated, please use either autodetect.hpp or generic.hpp" )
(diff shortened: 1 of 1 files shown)


include/cucumber-cpp/defs.hpp
```diff
@@ -1,4 +1,8 @@
-#warning Use of defs.hpp is deprecated, please use either autodetect.hpp or generic.hpp
+#ifdef __GNUC__
+#warning "Use of defs.hpp is deprecated, please use either autodetect.hpp or generic.hpp"
```
paoloambrosio: Indent the `#warning` that is inside the if block

include/cucumber-cpp/defs.hpp
```diff
@@ -1,4 +1,8 @@
...
+#else
+#pragma message( "Use of defs.hpp is deprecated, please use either autodetect.hpp or generic.hpp" )
```
paoloambrosio: Indent the `#pragma` that is inside the else block

aallrd:
I updated my change with a 4 space indentation.

Issue #127, opened 2016-11-04
StepManager.hpp causes C4512 warning for each step definition in Visual Studio


mpjmpj:
This occurs on Visual Studio 2013, x64 (probably others).
StepInfo contains a const member variable.  The compiler can't generate an assignment operator for it.  This causes a warning.  It generates a warning for each  step definition.  

There are no problems with functionality.  Just annoying warnings.  A simple fix is to create a private definition of the assignment operator to tell the compiler to knock it off.  

```
[redacted file name].cpp(34) : see reference to function template instantiation 'int cucumber::internal::registerStep<Y::CukeObject16>(const std::string &,const char *,const int)' being compiled
[redacted]\src\cucumber-cpp\include\cucumber-cpp\internal\step/StepManager.hpp(142): warning C4512: 'cucumber::internal::StepInvoker<T>' : assignment operator could not be generated
9>          with
9>          [
9>              T=Y::CukeObject17
9>          ]
9>          [redacted]\src\cucumber-cpp\include\cucumber-cpp\internal\step/StepManager.hpp(177) : see reference to class template instantiation 'cucumber::internal::StepInvoker<T>' being compiled
9>          with
9>          [
9>              T=Y::CukeObject17
9>          ]
9>          [redacted file name](41) : see reference to function template instantiation 'int cucumber::internal::registerStep<Y::CukeObject17>(const std::string &,const char *,const int)' being compiled
```

paoloambrosio:
Any idea why this doesn't happen in VS 2015 (used by the CI build on AppVeyor)?

mpjmpj:
No, I don't.  I poked around cmake, and didn't see anything that might turn off that warning.   My only guess is that the cucumber-cpp build is not building a test that exercises the StepInvoker template class in the same way that I am.   

https://msdn.microsoft.com/en-us/library/hsyx7kbz.aspx
MS specifically calls out that if you have a const member variable in a base class, it will not be able to generate an assignment operator for the derived class.  StepInvoker is the only class that inherits from StepInfo, which has a const member.  I'd think if it were being used, it would throw that warning. 

I'll try to poke around a bit more and see if I can generate a test that throws the warning.  

Is there a guide for the contributing to cucumber cpp?  I'm trying to figure out how to build and run things just like the build would do.  I might just be missing it though.

paoloambrosio:
[...]

PR #128, opened 2016-11-04
Fix issue #127 Assignment operator not generated

mpjmpj:
Added a private declaration of the assignment operator to StepInfo.

### DIFF
--- a/include/cucumber-cpp/internal/step/StepManager.hpp
+++ b/include/cucumber-cpp/internal/step/StepManager.hpp
@@ -105,6 +105,9 @@ public:
     step_id_type id;
     Regex regex;
     const std::string source;
+private:
+    StepInfo& operator=(const StepInfo& other);
+    
 };
 
 


mpjmpj:
Let me know if I did something wrong with the pull request.  This is my first time contributing to something on github, so I'm still learning the ropes.  Thanks!

paoloambrosio:
Thanks for this pull request. Don't worry about doing it wrong :-) The main rule is to follow the same code style as the rest of the code and write tests (TDD if possible). If you need to ask anything we have also a Gitter channel.

I cannot see the same warning on VS2015 (not 2013) run by AppVeyor.

My only concern with your implementation is that the purpose of that method is not clear, but perhaps it's just me not being a decent C++ developer. What about private inheritance from `boost::noncopyable` (that AFAIK for C++03 has your same implementation)?

mpjmpj:
I haven't worked in many c++ shops, but my feeling that this is a common practice.  If copy constructors and assignment operators aren't defined, the compiler generates them.  If you create a class where those generated functions are invalid (if you have a const member variable or a member variable that is a reference), you define them as private to signal to the compiler to skip the generation.  In one of the latest revisions of c++ they included a new syntax `thing& operator=(const thing&) = delete`.  But I figured you probably didn't want to break compatibility with older compilers if you didn't have to.

I haven't looked into boost::noncopyable (I will try to) but my own bias would be to avoid adding dependencies when there is a simple way to accomplish the same thing.  I suppose the benefit is that it self documenting - but again, it feels like a common pattern to me as-is.  All that said, I'm playing in your sandbox, so I'm happy to do whatever.

I do feel that there is something missing from the testing.  I don't get this warning when building the cucumber-cpp project on its own (even in vs2013).  I only get it when incorporating it into my project's tests.  I feel that if those classes were invoked by the cucumber-cpp unit tests, they'd generate the warning.  I just haven't had time to really look into it as we have a bit of a deadline at work.  I still plan on trying to however - it might just be next week.

paoloambrosio:
We already have a dependency on Boost for `boost::noncopyable`, but I'm OK merging this anyway. I'll do it ASAP.

PR #129, opened 2016-11-16
cucumber-cpp support for cgreen test framework.

shnewto:
As a user I want to use cgreen as an alternative test framework to gtest
and boost in order to test native C complete with mocking capability.

cucumber-cpp now recognizes and harnesses cgreen as a testing framework.
The cgreen reporting is intercepted and presented on the failure of a
cucumber
test.

An example (FizzBuzz) is written as an example of a testable C progam
and
is included in the examples directory.

Unit tests for the cgreen driver are modeled after the boost and gtest
drivers and are passing.


snewt:
Cgreen repo lives [here](https://github.com/cgreen-devs/cgreen).

paoloambrosio:
Thanks for submitting this PR! :-) I've been quite busy to do an in-depth review but I'll do it ASAP.

In the meantime some observations that will keep you busy:
 - The CI build doesn't seem to include it
 - The code style is sometimes inconsistent with the rest of the code base (e.g. mixed camelCase and underscore_case, "char*" VS "char *", brackets before or after a new line, ...)
 - No copyright information please
 - No spaces at the beginning of files

snewt:
Thanks for the update! 

I've addressed the following in this PR:
- CgreenDriver and FizzBuzz example integrations into the CI builds.
- Consistency in code style.

Maintaining the copyright information is a hard requirement for me though. 

Although the project reports "Copyright (c) 2010 Paolo Ambrosio"
my understanding is that copyright of contributors is still implicit 
and recorded by Github's record of project participants.

Unfortunately the role of PolySync Technologies, Inc. is not
reflected in that report and it needs to be somehow. I'd be 
willing to put some time into a project contributor's page if 
that's a more palatable option.

paoloambrosio:
No other source file in the code base is polluted with copyright information, and we are not going to start now. I'm more than happy to accept PRs from "PolySync Technologies, Inc.", but we'll have to reach an agreement before this PR is considered for merging.

The license chosen is among the most permissive in the OSS world. I'm surprised to receive this request from a company that has [open-sourced software](https://github.com/PolySync/OSCC). Does it mean that the company would accept in its project a contribution assigning copyright to me?

Contributors are recorded in the Changelog file when merging each PR, as well as being in the Git logs. The role of "PolySync Technologies, Inc." is contributing with that PR, and I have no problems attributing the change to the company instead of you. If this is agreeable, we don't need to discuss it anymore :-)

If that is not enough, I think we should discuss this in the https://gitter.im/cucumber/contributors channel and hear the opinion of the others core contributors.

aslakhellesoy:
Each repository under the cucumber GitHub organisation should have one and only one legal entity as copyright holder.

Most of the repositories under the "Cucumber" GitHub organisation have an MIT license where the copyright holder is Cucumber Ltd. For historical reasons the copyright holder of this repo is Paolo Ambrosio.

Cucumber Ltd (who owns the Cucumber GitHub organisation) will not allow more than one copyright holder for Cucumber Ltd repos, and I recommend the same policy for repositories like this one.

If a project wants to stray from this recommendation they would have to fork and move outside the Cucumber organisation, something I'm sure @paoloambrosio isn't planning on ;-)

So yes, these `Copyright PolySync Technologies, Inc.` lines must be removed before this can be merged.

snewt:
Copyright info removed from PR.

aslakhellesoy:
Thanks for your understanding @snewt - and glad you're happy with an honorable mention in the changelog!

And thanks for your contribution.

Cheers,
Aslak

paoloambrosio:
I'm slowly going through this PR (apologies for that).

It took me some time to realise that it doesn't work with the released version of Cgreen, that doesn't even come with CMake config files. I know, I missed that your contribution requires "Cgreen 1.1.0 or later"... but in my defence that version does not exist!

Also Cgreen's CMake config in the master branch doesn't export the library's full path, making my local build fail since I keep Cucumber-CPP's dependencies in a separate directory (solved exporting the `LIBRARY_PATH` variable). I will consider contributing back to cgreen the more standard CMake config file that I just wrote.

I also probably found the reason why you didn't add cgreen to the AppVeyor build for Windows: cgreen does not compile with Visual Studio, and indeed they don't build on Windows. Again, spent time solving a few issues but I gave up after a while (I don't even like Windows!).

I like that you added the FizzBuzz example to test a pure C application. I would also create a CgreenCalculatorSteps.cpp file in examples/Calc/features/step_definitions to show how to use it with C++. That directory in my view should contain one file per test framework supported.

More to follow after I take a look at the actual code ;-)

CMakeLists.txt
```diff
@@ -100,6 +100,13 @@ if(NOT CUKE_DISABLE_GTEST)
 endif()
 
 #
+# cgreen
+#
+if(NOT CUKE_DISABLE_CGREEN)
+    find_package(cgreen QUIET)
```
paoloambrosio: `QUIET` is not supported by `cgreen-config.cmake`, and even if it was all other `find_package` in the project do not use it

examples/FizzBuzz/features/step_definitions/FizzBuzzSteps.cpp
```diff
@@ -0,0 +1,43 @@
+#include <cgreen/cgreen.h> // general unit testing
+#include <cgreen/mocks.h> // mocking functionality
+
+#include <cucumber-cpp/autodetect.hpp>
+#include <cucumber-cpp/internal/CukeEngine.hpp>
```
paoloambrosio: This should not be included

include/cucumber-cpp/internal/drivers/CgreenDriver.hpp
```diff
@@ -0,0 +1,38 @@
+#ifndef CUKE_CGREENDRIVER_HPP_
+#define CUKE_CGREENDRIVER_HPP_
+
+#include <cgreen/cgreen.h> // general unit testing
```
paoloambrosio: Is the comment needed?

tests/integration/drivers/CgreenDriverTest.cpp
```diff
@@ -0,0 +1,63 @@
+#include <cgreen/cgreen.h> // general unit testing
```
paoloambrosio: Same as before

tests/integration/drivers/CgreenDriverTest.cpp
```diff
@@ -0,0 +1,63 @@
...
+};
+
+class CgreenDriverTest : public DriverTest {
+public:
+    virtual void runAllTests() {
+        stepInvocationInitsBoostTest();
+        DriverTest::runAllTests();
+    }
+private:
+    void stepInvocationInitsBoostTest() {
```
paoloambrosio: Not boost :-)

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+    va_list arguments;
+    memset(&arguments, 0, sizeof(va_list));
+
+    cgreenInterceptor.cgreenReporter->show_fail(
+            cgreenInterceptor.cgreenReporter,
+            __FILE__,
+            __LINE__,
+            "currentTest",
+            arguments);
```
paoloambrosio: I'm struggling to follow what this is doing, but I don't think `__FILE__` and `__LINE__` can work here. Given that it's not in a macro, I believe they would be `CgreenDriver.cpp` and `145`.

CMakeLists.txt
snewt: `QUIET` is a cmake directive, it allows the cmake step to succeed if the `find_package` function fails to find cgreen. Without the QUIET directive cmake can fail there. I'm happy to remove it if you want cmake to fail if the package is not found.

src/drivers/CgreenDriver.cpp
snewt: Hmm, I'm unclear what the concern here is. `__FILE__` and `__LINE__` are standard predefined macros that report source file name and line number, respectively. They are used here to give the cgreen reporter the info it requires for the proper output.

CMakeLists.txt
paoloambrosio: `QUIET` suppresses both success and failure messages. For consistency with the other `find_package` calls, until , I suggest adding a more standard `Findcgreen.cmake`.

Feel free to pull [this commit](https://github.com/paoloambrosio/cucumber-cpp/commit/973e44406058a26702a9bdde8601761953995f05) on my personal repository that adds it on top of your PR.

src/drivers/CgreenDriver.cpp
paoloambrosio: I can't understand what this is doing:
```
    va_list arguments;
    memset(&arguments, 0, sizeof(va_list));

    cgreenInterceptor.cgreenReporter->show_fail(
            cgreenInterceptor.cgreenReporter,
            __FILE__,
            __LINE__,
            "currentTest",
            arguments);
```
Even if I remove the entire block, `CgreenDriverTest` passes and the FizzBuzz example works (even failures are reported correctly if I introduce errors in the code).

snewt:
@paoloambrosio good catch on that `show_fail` function, my impression was that I needed to call it to fill the runner output with the correct strings but I tested it as well and it seems not to be the case! I also merged in you're find package solution, thanks!

snewt:
Recently resolved conflicts in the toml and added some Calc / C++ examples. Any more concerns to address?

konserw:
Does cgreen support building with cmake? couldn't we use externalproject for it?

snewt:
@konserw Cgreen does build with CMake. Your suggestion is a possibility but I followed the Boost / Qt lead, require a Cgreen install and use `find_package` to do the legwork. Is that impractical?

.travis.yml
```diff
@@ -22,13 +22,18 @@ matrix:
       compiler: gcc
       env: GMOCK_VER=1.8.0 VALGRIND_TESTS=ON
 
+before_script:
+  - git clone https://github.com/cgreen-devs/cgreen.git
```
paoloambrosio: Given that after several months there is still no sign of a 1.1.0 release, I think we should fix our build to a specific git hash. I want to avoid our build failing when someone makes incompatible changes to Cgreen.

konserw:
> @konserw Cgreen does build with CMake. Your suggestion is a possibility but I followed the Boost / Qt lead, require a Cgreen install and use find_package to do the legwork. Is that impractical?

What I had in mind is approach simmilar to what we have for gtest/gmock, especially since there is no cgreen package available for ubuntu (which we use in CI), i.e. if cgreen is available in system use that, otherwise use externalproject to clone and build cgreen for cucumber.
Of course it is only my opinion, what do you think @paoloambrosio ?

paoloambrosio:
I'm not a fan of externalproject so I'm happy to merge this PR with the current CMake file.

CMakeLists.txt
```diff
@@ -141,6 +141,60 @@ if(VALGRIND_TESTS)
...
+# cgreen
+#
+
+if(NOT CUKE_DISABLE_CGREEN)
+    find_package(cgreen)
+endif()
+
+#
+# Valgrind
+#
```
paoloambrosio: The build errors are due to this section being duplicated (suspect during rebasing conflict).

paoloambrosio:
For your next PR, I suggest rebasing earlier instead of merging from master. Apologies for not spotting it earlier: it could have saved you a headache. You should be able to solve that problem with the following instructions.

Start an intereactive rebase from master: `git rebase -i master` and set the second commit for editing:
```
pick ed59fec cucumber-cpp support for cgreen test framework.
edit 883984f Added Cgreen Calc and CalcQt examples.
pick 60a367f checkout known good cgreen state before build
```

Unstage changes: `git reset HEAD^`

Edit `CMakeLists.txt` removing the second Valgrind section.

Commit the changes with `git commit -a` and the original message:
```
    Added Cgreen Calc and CalcQt examples.
    
    In order to show how the Cgreen framework can be used with C++,
    Cgreen tests, modeled after the Boost tests were written
    for the Calc and CalcQt examples.
```

Continue rebasing: `git rebase --continue`

snewt:
@paoloambrosio thanks for the pointers, rebase rather that merge is a workflow I need to master. Also, thanks for putting the time into spotting the duplicate code / bad rebase conflict resolution.

snewt:
@paoloambrosio was experimenting with a fresh VM tonight and realized that the cgreen driver is no longer found on Linux, the OS X builds on Travis run the cgreen examples but the linux builds do not. I think I've spotted the culprit (cgreen_FOUND vs CGREEN_FOUND), will push later tonight

snewt:
Fixed the`find_package(cgreen)` issue, examples and tests are running on OS X and Linux builders now but it looks like I have a memory leak to pin down.

snewt:
update, I have a PR up on the cgreen repository to fix the leak that is causing my the cgreen valgrind tests to fail.

snewt:
Okay, the Cgreen team has merged my PR. Had to add a few lines to the CgreenDriver source after bumping the version to account for some new error message output but it looks like we actually have a known good state in the hash we checkout in the travis file.

snewt:
hmm, a couple AppVeyor builds are failing with a FindBoost.cmake error, not sure how to tackle that one. @paoloambrosio I'm definitely open to suggestions.

snewt:
It looks like the variables in the appveyor.yml passed as arguments to cmake don't like `\` in paths. In a test branch on my fork I modified all the appveyor.yml variables to use `/` in paths instead, see diff [here](https://github.com/cucumber/cucumber-cpp/compare/master...snewt:appveyor-build-tests?expand=1#diff-180360612c6b8c4ed830919bbb4dd459) and appveyor success [here](https://ci.appveyor.com/project/snewt/cucumber-cpp/build/appveyor-build-tests-ci-3). The only diff on that branch and this PR is that appveyor.yml file. Seems out of the scope of this PR and very weird that it's only causing an issue now but I can add those changes to this PR if you're interested @paoloambrosio ?

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,170 @@
...
+int CukeCgreenInterceptor::cgreenPrinter(const char* format, ...) {
+    char buffer[10000] = { 0 };
+
+    va_list argPtr;
+
+    va_start(argPtr, format);
+
+    vsprintf(buffer, format, argPtr);
+
+    if(not is_blacklisted(buffer)) {
```
muggenhor: Please just use the `!` operator instead of `not`. Very few C++ developers use the "word" form of the logical operators and it's not even available in C (the preprocessor doesn't count). Using the more common form (`!`) helps readability.

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,170 @@
...
+        cgreenOutput += "\n\n";
+        cgreenOutput += buffer;
+        cgreenOutput += "\n\n";
+    }
+
+    va_end(argPtr);
+
+    std::string blackBoxInfo = "\"initCgreenTest\"";
```
muggenhor: make the type `static const char blackBoxInfo[] = ...` instead. This code causes a heap allocation and memcpy on every invocation of this function.

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,170 @@
...
+    std::size_t position = cgreenOutput.find(blackBoxInfo);
+
+    if(position != std::string::npos) {
+        cgreenOutput.resize(position);
```
muggenhor: Please use `.erase(position)` instead. It does the same thing, but it describes your intent better in this case (erasing "initCgreenTest" and everything that comes after).

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    return cgreenOutput.length();
+}
+
+bool CukeCgreenInterceptor::is_blacklisted(const char* input) {
+    const int blacklist_size =
+            sizeof(blacklist)
+                    / sizeof(blacklist[ 0 ]);
+
+    if(strlen(input) < 1) {
```
muggenhor: 1. I find `strlen(input) < 1` more difficult to read than `strlen(input) == 0` or even `input[0] == '\0'`
2. Instead of special-casing the empty string I suggest adding the empty string to your blacklist array instead (adding it to the top is probably the best for performance).

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,170 @@
...
+        return true;
+    }
+
+    for(int idx = 0; idx < blacklist_size; ++idx) {
+        if(strncmp(input, blacklist[idx], strlen(blacklist[idx])) == 0) {
```
muggenhor: Unless you are looking for prefixes instead of exact matches, you should use `strlen(blacklist[idx]) + 1` to ensure the NUL-byte gets compared too.

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+
+    return false;
+}
+
+void CgreenStep::setReporterPrinter(
+        TestReporter* reporter,
+        TextPrinter* newPrinter) {
+    TextMemo* memo = (TextMemo*)reporter->memo;
```
muggenhor: please use `reinterpret_cast` instead.

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+    if(not initialized) {
+        initCgreenTest();
+    } else {
+        cgreenInterceptor.reset_reporter();
+    }
+
+    currentTestBody = boost::bind(&CgreenStep::body, this);
+
+    suite_result = run_single_test(
```
muggenhor: Please declare and initialize this variable here. I.e.:
```cpp
const int suite_result = ...
```

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,170 @@
...
+
+using namespace cgreen;
+
+namespace cucumber {
+namespace internal {
+
+
+bool CgreenStep::initialized = false;
+
+boost::function< void() > currentTestBody;
```
muggenhor: Why don't you make this a `static` class member instead?

tests/integration/drivers/CgreenDriverTest.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+class CgreenDriverTest : public DriverTest {
+public:
+    virtual void runAllTests() {
+        stepInvocationInitsCgreenTest();
+        DriverTest::runAllTests();
+    }
+private:
+    void stepInvocationInitsCgreenTest() {
+        std::cout << "= Init =" << std::endl;
```
muggenhor: Unless you really want .flush() to be called here I suggest you just terminate the string with `"\n"` instead of streaming in `std::endl`. `std::endl` does nothing more than streaming in `"\n"` followed by calling `.flush()`.

FYI: some people think it takes care of converting to the current OS' end-of-line character sequence: it doesn't, that's what the underlying stream itself already does for `"\n"` (exactly like _not_ passing the "b" flag to fopen() does for C, in fact: on GNU's libstdc++ it uses the C library).

travis.sh
```diff
@@ -31,3 +32,7 @@ if [ -f $BOOST ]; then
     cucumber examples/Calc
     wait
 fi
+if [ -f $CGREEN ]; then
+    $CGREEN >/dev/null &
+    cucumber examples/FizzBuzz
```
muggenhor: Please add a `wait` call after `cucumber`. That makes sure that to wait for `$CGREEN` to terminate (and to use its exit code).

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,170 @@
...
+    cgreenReporter->exceptions = 0;
+}
+
+int CukeCgreenInterceptor::cgreenPrinter(const char* format, ...) {
+    char buffer[10000] = { 0 };
+
+    va_list argPtr;
+
+    va_start(argPtr, format);
+
```
muggenhor: I see this a lot: empty lines between related code. Could you try to remove some of that?

In this case, for example, I wouldn't expect to see a single empty line between the `va_list` declaration and the `va_end` call, as it's all related and logically doing one thing.

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,170 @@
...
+    ~CukeCgreenInterceptor();
+
+private:
+    static std::string cgreenOutput;
+    static bool is_blacklisted(const char* input);
+};
+
+static CukeCgreenInterceptor cgreenInterceptor;
+
+std::string CukeCgreenInterceptor::cgreenOutput = std::string();
```
muggenhor: Please remove the copy-initialization. In C++ all objects of types that have a default constructor (which is almost all non-POD types) _will_ get it called when no initializer is specified (so no need to worry about uninitialized memory here).

src/drivers/CgreenDriver.cpp
snewt: We are looking for prefixes here.

snewt:
@muggenhor thanks for all the great feedback, the changes you requested are well justified. I really appreciate you taking the time to explain your reasoning.

src/drivers/CgreenDriver.cpp
snewt: I think that since the blacklist isn't looking for exact / full strings but only the first bit of strings to exclude from output, including the empty string in the blacklist would be problematic.

snewt:
@muggenhor I've tried to address your requested changes. Thanks again for the comments, I learned some things implementing them. I'm unsure about my README.md wording still, I don't know the best way to indicate a supported cgreen version when there hasn't been an actual version in some time. I've been using cgreen and cucumber-cpp in tandem for some time now and feel like even at the stage of development cgreen is currently at it is still a valuable tool. It didn't seem like a commit hash in the README was appropriate though.

snewt:
Also, it appears the `/` v `\` in variables in the appveyor builds is still causing problems.

snewt:
I kicked off another appveyor build of branch that differs from this one only in the `appveyor.yml` where `\` in variable paths have been changed to `/`. You can check out the file diff [here.](https://github.com/cucumber/cucumber-cpp/compare/master...snewt:appveyor-build-tests#diff-180360612c6b8c4ed830919bbb4dd459) and the build status [here.](https://ci.appveyor.com/project/snewt/cucumber-cpp/build/appveyor-build-tests-ci-6). Let me know what you'd like me to do regarding those failures on this PR.

travis.sh
snewt: [Added](https://github.com/cucumber/cucumber-cpp/pull/129/files#diff-f9a62833538ab14fff849c2be16ab986)

snewt:
@muggenhor regarding the request to remove the blanketed `include_directories(${CGREEN_INCLUDE_DIR})`. I removed that from the `Findcgreen.cmake` file but then reverted the CMakeLists.txt files that build cgreen dependent targets to use `target_include_directories(target_name PRIVATE ${CGREEN_INCLUDE_DIR})`. We don't have to do the same for Boost or Gtest because their headers are in a system visible location right? If you or @paoloambrosio have an approach you'd prefer to the one I've taken I'm certainly open to suggestions.

examples/FizzBuzz/src/FizzBuzzReporter.h
```diff
@@ -0,0 +1,11 @@
+#ifdef __cplusplus
```
snewt: [Added](https://github.com/snewt/cucumber-cpp/blob/2763ecba245c9acc48e3beab2a339c1bbfd58946/examples/FizzBuzz/src/FizzBuzzReporter.h)

src/CMakeLists.txt
```diff
@@ -24,6 +24,10 @@ if(Boost_UNIT_TEST_FRAMEWORK_FOUND)
     list(APPEND CUKE_SOURCES drivers/BoostDriver.cpp)
 endif()
 
+if(CGREEN_FOUND)
+    list(APPEND CUKE_SOURCES drivers/CgreenDriver.cpp)
```
muggenhor: Just like is done for GTest (few lines above): you need to ensure Cgreen gets linked to, like this:
```cmake
    list(APPEND CUKE_DEP_LIBRARIES Cgreen::Cgreen)
```

snewt:
@muggenhor ah thanks, I'd misremembered.

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,159 @@
...
+static const char* const blacklist[] = {
+        "Running",
+        "Completed",
+        "Failure",
+        "currentTest",
+        "\n",
+        "/"
+};
+
+class CukeCgreenInterceptor {
```
muggenhor: This class smells: it's basically a singleton (the single `static` instance of this class) with every member being a singleton too (`static` member variables)! I suspect the quality of your destructor (which to me looks like it will trigger NULL pointer dereferences) is the result of this.

Please:

1. do without any singleton at all, or;
2. don't nest singletons and make the top-level one closer to the [singleton design pattern](https://en.wikipedia.org/wiki/Singleton_pattern) with lazy initialization (use a smart pointer for the instance pointer to ensure the destructor actually gets called).

snewt: The project doesn't seem to enforce c++11 so I can't assume std::unique_ptr is available.

src/drivers/CgreenDriver.cpp
```diff
@@ -0,0 +1,159 @@
...
+}
+
+const InvokeResult CgreenStep::invokeStepBody() {
+    if(!initialized) {
+        initCgreenTest();
+    } else {
+        cgreenInterceptor.reset_reporter();
+    }
+
+    cgreenInterceptor.currentTestBody = boost::bind(&CgreenStep::body, this);
```
snewt: The `body()` function of the BasicStep class is protected.

src/drivers/CgreenDriver.cpp
muggenhor: Correct, you should assume only C++98, as we currently still are able to build with just that. You don't need `std::unique_ptr` for a singleton though, that just makes it slightly easier to implement.

snewt: @muggenhor got it. Well I took a stab at implementing the lazy evaluate singleton you requested without the unique_ptr. I'm not dynamically allocating memory for the singleton itself though there are fields in it that are allocated and freed in its destructor. My assumption is that that memory is being successfully managed since the Valgrind tests pass. Let me know what your thoughts are.

snewt:
I've been unable to reproduce the BoostCalculatorQtSteps failure that's breaking the Travis/Linux builds but am investigating. I triggered a build of the current state of master from my fork and it seems to be broken there too.

konserw:
It's something with travis image itself - see #170 #169 I'm not sure what can we do about that

konserw:
@snewt - FYI if you rebase now onto master travis build should pass (issue has been worked-around in master)

paoloambrosio:
Don't really like merges from master. Rebase is your friend, like @konserw suggested.

travis.sh
```diff
@@ -73,15 +76,27 @@ done
 for TEST in \
     build/examples/CalcQt/GTestCalculatorQtSteps \
     build/examples/CalcQt/BoostCalculatorQtSteps \
+    build/examples/CalcQt/CgreenCalculatorQtSteps \
 ; do
     if [ -f "${TEST}" -a -n "${DISPLAY:-}" ]; then
-        "${TEST}" 2> /dev/null &
```
konserw: Don't change this line - it will produce a lot of garbage if you change it

travis.sh
```diff
@@ -92,4 +107,7 @@ if [ -f "${TEST}" ]; then
     wait %
 fi
 
-killXvfb
```
konserw: also don't change this one -

snewt:
@konserw thanks for pointing those issues out. Looks like my build is still failing though. Any more suggestions? I experimented with another branch and the failure seems to be rooted in this block of the `travis.sh` 

```
SOCK=cucumber.wire.sock
  TEST=build/examples/FeatureShowcase/FeatureShowcaseSteps
 if [ -f "${TEST}" ]; then
     echo "unix: ${SOCK}" > examples/FeatureShowcase/features/step_definitions/cucumber.wire
     "${TEST}" --unix "${SOCK}" > /dev/null &
     cucumber examples/FeatureShowcase
     wait %
 fi
```

jermus67:
Hallo,

Despite really appreciating the effort the @shnewto has put into this PR, I also think it is not a good direction to add support for various (unit) test frameworks to cucumber-cpp. This would make cucumber-cpp dependent on each and every test framework it is embedded in, while it should be the other way around: the test of the user of cucumber-cpp should be depending on both test framework it runs on and cucumber-cpp, the dependency for cucumber-cpp towards the test framework should be injected by the user of the test by means of a - by cucumber-cpp - well defined interface (kind of as describe in issue #154 ).

This PR will be closed without being merged into main (besides, this PR seems to be _really_ out of date).

However, contributors to this PR are more than welcome to participate in the design of a plugin / dependency injection system for test drivers in cucumber-cpp.

Kind regards,
Jeroen Kouwer

PR #109, opened 2016-03-17
Ninja build

konserw:
Hi once again ;)
I've added compatibility with [Ninja](https://ninja-build.org/) build system, and it builds fine except travis OSX.
I hope someone ( @MartinDelille I'm looking at you ;) ) could help me with that problem, please see [my travis build](https://travis-ci.org/konserw/cucumber-cpp/jobs/116736676#L165) for details.
Any other comments would also be welcome ;)

PR #130, opened 2016-12-02
Find g mock

meshell:
This pull request fixes the cmake files for building with MinGW under windows.
Further it modifies the FindGMock.cmake file to only download and build googlemock and googletest when not already available. This is IMHO helpful, because usually you already have googletest and/or googlemock installed for your unit tests.


konserw:
Why have you inserted all those spaces in if() and else() ?

meshell:
@konserw I edited the files with CLion and the spaces sneaked in when I reformated the file. I didn't recognize it because whitespace differences are turned of in my diff tool. If this is  a problem I will reformat the files.

src/CMakeLists.txt
```diff
@@ -35,3 +35,12 @@ endif()
...
 add_library(cucumber-cpp STATIC ${CUKE_SOURCES} main.cpp)
+
+if(WIN32 AND NOT MSVC)
+    target_link_libraries(cucumber-cpp-nomain
+        ws2_32
+        )
+    target_link_libraries(cucumber-cpp
+        ws2_32
+        )
+endif(WIN32 AND NOT MSVC)
```
meshell: Actually I don't know if other compilers need this library as well. i only have MinGW and MSVC available on my windows machine. I know that MinGW requires it, but MSVC doesn't. However I think it would also work when the library is added on MSVC also. I will check that.

paoloambrosio: I'd be happy with `if(MINGW)` (if I remember correctly) unless further evidence suggests that it's needed for all `WIN32` compilers that are `NOT MSVC`.

konserw:
@paoloambrosio - Looks good to me, but I'm not CMake expert :)
@meshell - Could you add Mingw build to appveyor CI ?

meshell:
@konserw I will try to add Mingw to appveyor CI. But I will do it in a new pull request. Is this ok?

paoloambrosio:
Adding MinGW to AppVeyor makes sure that the build works on MinGW without testing is manually.
Doing it as a separate PR would be like adding tests for code changes as a separate PR.

I understand it's going to be quite painful and involve a lot of commits. What I suggest is to experiment with it in a separate branch in your repository, and add it to this PR when you are reasonably confident it will work :-)

meshell:
@paoloambrosio: Ok I will do that. 
That problem are the boost libraries. I'm afraid, that the boost libraries must be built as part of the project. I already did that on an other project, but I don't know if it is the right approach for cucumber-cpp. I could do it similar to the approach done for the googlemock library, i.e. download and build in a FindBoost.cmake file, when not already available. However this means probably to copy the existing FindBoost.cmake provided with cmake. The other approach would be to have a separate target to build external dependencies. This target must than be build before everything else. What do you think, which approach I should try?

paoloambrosio:
If Boost is not compiled for MinGW on AppVeyor, I'd be OK without a MinGW build. Unfortunately this also means that any change might silently break it.

I'm not a big fan of the build script downloading and compiling dependencies (like the current GoogleTest implementation). It is also not how the standard CMake scripts work.

IMO anything related only to the CI build should be in the CI configuration (`travis.yml`, `appveyor.yml`) and scripts used by those (`travis.sh`). I would even move build steps in a `scripts/` directory and have `travis.yml` and `appveyor.yml` call them instead of having code and configuration mixed.

konserw:
After quick google search i think for we could install boost for mingw via MSYS2 pacman

meshell:
I had to do some minor changes in some CMake files as well, but finally I managed to add Mingw build to appveyor.

muggenhor:
This PR duplicates #109 btw.

meshell:
@muggenhor I can't see the duplication. Could you please explain.

CMakeLists.txt
```diff
@@ -81,21 +83,22 @@ endif()
...
     if(GMOCK_FOUND)
         set(CUKE_GMOCK_LIBRARIES
-            ${GTEST_BOTH_LIBRARIES}
-            ${GMOCK_BOTH_LIBRARIES}
-            ${CMAKE_THREAD_LIBS_INIT}
-        )
+            GTest::GTest
+            GMock::GMock
+            GMock::Main
+            )
```
meshell: @muggenhor I totally agree. I would also do it differently and would not use this variables at all. However I didn't want to change too many parts of the existing CMake files in this PR.

muggenhor:
@meshell no duplication in code. I just meant that this PR seems to try to accomplish the same thing, so that they're mutually exclusive.

CMakeLists.txt
muggenhor: I did something like that a while back as an experiment: https://github.com/muggenhor/cucumber-cpp/compare/master...link-depends-only, and it's big enough a change that it definitely belongs in a different PR. But it does depend on doing the same kind of change that you're doing happening first.

cmake/modules/FindGMock.cmake
```diff
@@ -113,172 +117,329 @@ function(GTEST_ADD_TESTS executable extra_args)
...
+                LOG_CONFIGURE ON
+                LOG_BUILD ON
+                CMAKE_ARGS
+                    -Dgtest_disable_pthreads=${MINGW}
+                    ${GTEST_CMAKE_ARGS}
+                DEPENDS gtest
+            )
+
+
+            add_dependencies(GTest::GTest gtest)
```
muggenhor: If you specify `BUILD_BYPRODUCTS` you don't need these custom dependency declarations at all (as long as you use one of those byproducts in the library's `IMPORT_LOCATION`).

cmake/modules/FindGMock.cmake
```diff
@@ -113,172 +117,329 @@ function(GTEST_ADD_TESTS executable extra_args)
...
+        message(STATUS "Building Gmock / Gtest from dir ${GMOCK_SRC_DIR}")
+        externalproject_add(
             gmock
-            GIT_REPOSITORY "https://github.com/google/googlemock.git"
-            GIT_TAG "release-${GMOCK_VER}"
+            SOURCE_DIR ${GMOCK_SRC_DIR}
             PREFIX ${GMOCK_ROOT}
             INSTALL_COMMAND ""
-            LOG_DOWNLOAD ON
+            LOG_DOWNLOAD OFF
```
muggenhor: Why don't just leave it as is? It's not displayed on the terminal anyway, but if something goes wrong you'll now have a log file to look at. This is pretty relevant in e.g. containerized builds where sometimes people mess up with the networking.

cmake/modules/FindGMock.cmake
```diff
@@ -113,172 +117,329 @@ function(GTEST_ADD_TESTS executable extra_args)
...
-        LOG_DOWNLOAD OFF
-        LOG_CONFIGURE ON
-        LOG_BUILD ON
-        CMAKE_ARGS ${GTEST_CMAKE_ARGS}
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(GTest DEFAULT_MSG GTEST_LIBRARY GTEST_INCLUDE_DIR GTEST_MAIN_LIBRARY)
+find_package_handle_standard_args(GMock DEFAULT_MSG GMOCK_LIBRARY GMOCK_INCLUDE_DIR GMOCK_MAIN_LIBRARY)
+
+include(CMakeFindDependencyMacro)
+find_dependency(Threads)
```
muggenhor: Just wondering here: what's the advantage of these two lines over the single: `find_package(Threads REQUIRED)` (which also creates the `Threads::Threads` target)? Is it maybe related to not being able to use `find_package` in a `FindXXX` module?

cmake/modules/FindGMock.cmake
```diff
@@ -113,172 +117,329 @@ function(GTEST_ADD_TESTS executable extra_args)
...
+    IMPORTED_LINK_INTERFACE_LANGUAGES "CXX"
+    IMPORTED_LOCATION "${GMOCK_LIBRARY}")
+
+if(GMOCK_INCLUDE_DIRS)
+    set_target_properties(GMock::GMock PROPERTIES
+        INTERFACE_INCLUDE_DIRECTORIES "${GMOCK_INCLUDE_DIRS}")
+endif()
+
+set_target_properties(GMock::Main PROPERTIES
+    INTERFACE_LINK_LIBRARIES "GMock::GMock"
```
muggenhor: Do you still want to link transitively to the non-main GMock on 1.7+ where the main version contains the same sources, just one more?

I would think that for 1.7 and beyond not specifying it on the link interface and instead only adding the same include directories as `GMock::GMock` has to the interface include dirs would be the clean thing to do.

src/CMakeLists.txt
```diff
@@ -35,3 +35,12 @@ endif()
 
 add_library(cucumber-cpp-nomain STATIC ${CUKE_SOURCES})
 add_library(cucumber-cpp STATIC ${CUKE_SOURCES} main.cpp)
+
+if(MINGW)
```
muggenhor: The same should work for MSVC too (it does at work). So I suggest you try with `if(WIN32)` instead.

meshell:
@muggenhor I do not agree. This PR fixes the cmake files for building with MinGW under windows and it modifies the FindGMock.cmake file to only download and build googlemock and googletest when not already available (this was actually the main goal of this PR but the MinGW stuff took over a bit). PR #109 however adds ninja support and does not change the FindGMock the same way, or does it?
Why do you think it tries to accomplish the same?

tests/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ if(GMOCK_FOUND)
     cuke_add_test(integration/HookRegistrationTest)
     cuke_add_test(integration/StepRegistrationTest)
     cuke_add_test(integration/TaggedHookRegistrationTest)
-    if (NOT MSVC)
+    if(NOT WIN32)
```
muggenhor: Have you tried running this with MinGW? Because I don't think there's anything to this test that should prevent it from running on Windows. (Just asking because I'm only noticing that this test is disabled for Windows just now).

cmake/modules/FindGMock.cmake
```diff
@@ -113,172 +117,329 @@ function(GTEST_ADD_TESTS executable extra_args)
...
+endif()
+
+find_path(GMOCK_INCLUDE_DIR gmock/gmock.h
+    HINTS
+    $ENV{GMOCK_ROOT}/include
+    ${GMOCK_ROOT}/include
+    )
+mark_as_advanced(GMOCK_INCLUDE_DIR)
+
+find_path(GTEST_INCLUDE_DIR gtest/gtest.h
```
muggenhor: I'm assuming you're searching the content of the `google-mock` Debian package here. Does this also work with the newer 1.8 `google-mock`/`googletest` packages (Ubuntu Trusty only provides 1.6)? Because I needed to do some hacking in the previous implementation of `FindGMock.cmake` to be able to use those newer packages.

muggenhor:
@meshell because the underlying cause why Ninja builds don't work is because of a missing dependency declaration from the libraries to the `ExternalProject_Add` targets. That's caused by using the `${GXXX_XXX_LIBRARIES}` variables (which contain filenames, not target names) instead of the created library targets. You're fixing that part of the cause. If you additionally add the `BUILD_BYPRODUCTS` declarations then using the filenames would also work (and you'd also not need `add_dependencies` for the target case).

cmake/modules/FindGMock.cmake
meshell: Usually in a project where cucumber-cpp is used, also a unitttest framework is used, often this might be googletest/googlemock. Therefore my intention was to be able to use this existing (already compiled) version of googlemock and googletest and not compiling it again. I have to confess, that I din't try to use installed versions, i.e. debian packages. I will check that. Thanks for the hint.

cmake/modules/FindGMock.cmake
```diff
@@ -113,172 +117,329 @@ function(GTEST_ADD_TESTS executable extra_args)
...
+if(EXISTS "${GTEST_LIBRARY}" OR EXISTS "${GTEST_LIBRARY_DEBUG}" AND GTEST_INCLUDE_DIR)
+    set(GTEST_LIBRARY_EXISTS ON)
+endif()
+
+if(NOT (${GMOCK_LIBRARY_EXISTS} AND ${GTEST_LIBRARY_EXISTS}))
+
+    include(ExternalProject)
+
+    if(MSVC)
+        set(Suffix ".lib")
```
meshell: Good point.

cmake/modules/FindGMock.cmake
```diff
@@ -113,172 +117,329 @@ function(GTEST_ADD_TESTS executable extra_args)
...
+                GIT_REPOSITORY "https://github.com/google/googletest.git"
+                GIT_TAG "release-${GMOCK_VER}"
+                PREFIX ${GMOCK_ROOT}
+                INSTALL_COMMAND ""
+                LOG_DOWNLOAD ON
+                LOG_CONFIGURE ON
+                LOG_BUILD ON
+                CMAKE_ARGS
+                    -Dgtest_disable_pthreads=${MINGW}
+                    ${GTEST_CMAKE_ARGS}
```
meshell: I didn't want to change the existing code done by @konserw in the PR I just wanted to extend it. I could do it, but IMO it would be better to do all this CMake refactorings in a seperate PR

cmake/modules/FindGMock.cmake
meshell: This is a mistake. I will fix that. Thanks

cmake/modules/FindGMock.cmake
meshell: Yes as far as I know you can not use find_package in FindXXX modules. This FindGMock.cmake is an adapted version of a recent FindGTest.cmake file plus the code from @konserw and this how its done in the FindGTest.cmake.

src/CMakeLists.txt
meshell: Please see the comments by @paoloambrosio for that.

tests/CMakeLists.txt
meshell: Yes I tried to run it and it failed. I suppose for the same reason as when built with MSVC, but I didn't investigate further. Maybe @paoloambrosio knows why this test fails (under MSVC)

paoloambrosio: It's been there for ages and I tried several times to understand why this fails on Windows but without success. I remember it crashes but can't remember how to be honest :-) Maybe we should create an issue on GitHub and put the link in a comment to that if statement.

appveyor.yml
```diff
@@ -18,11 +42,15 @@ install:
 - gem install bundle
 - bundle install
 - bundle env
+- if "%build%"=="mingw" set PATH=%MINGW_ROOT%\bin;C:\msys64\usr\bin\;%PATH%
+- if "%build%"=="mingw" set MSYSTEM=%MSYSTEM%
```
konserw: I think this line does nothing (assigns variable to itself)

cmake/modules/FindGMock.cmake
konserw: I agree.

paoloambrosio:
Once @konserw's latest comment is addressed and commits are squashed (it doesn't need to be one single commit but 10 is a bit too much) and the branch rebased (to run the CI on top of the latest master), I think it can be merged.

Well done on the collaborative effort! 👍

cmake/modules/FindGMock.cmake
meshell: This is how it's done in a recent FindGTest.cmake provided with CMake. That's why I didn't consider removing it.

meshell:
@paoloambrosio I removed the lines mentioned by @konserw and rebased it. It is ready to merge.
Thanks.

paoloambrosio:
Thanks everyone! @meshell for the great PR, @muggenhor and @konserw for the reviews

PR #131, opened 2016-12-13
Add support for using ephemeral ports

muggenhor:
Ephemeral ports are used when binding to port '0'. The operating system
then automatically selects a (non-zero) port for you that's unused at
the time. This is great to prevent conflicts.

This commit extends the wire server such that it reports the actual port
number we've bound to, even when using ephemeral ports. This is useful
to prevent port conflicts on CI systems running many tests in parallel.

### DIFF
--- a/src/connectors/wire/WireServer.cpp
+++ b/src/connectors/wire/WireServer.cpp
@@ -18,6 +18,11 @@ void SocketServer::listen(const port_type port) {
     acceptor.listen(1);
 }
 
+SocketServer::port_type SocketServer::listenPort() const {
+  const tcp::endpoint ep(acceptor.local_endpoint());
+  return ep.port();
+}
+
 void SocketServer::acceptOnce() {
     tcp::iostream stream;
     acceptor.accept(*stream.rdbuf());
--- a/travis.sh
+++ b/travis.sh
@@ -18,8 +18,10 @@ BOOST=build/examples/Calc/BoostCalculatorSteps
 if [ -f $GTEST ]; then
     $GTEST >/dev/null &
     cucumber examples/Calc
+    wait
 fi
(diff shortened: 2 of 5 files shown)


travis.sh
```diff
@@ -18,8 +18,10 @@ BOOST=build/examples/Calc/BoostCalculatorSteps
 if [ -f $GTEST ]; then
     $GTEST >/dev/null &
     cucumber examples/Calc
+    wait
```
muggenhor: Without this it's possible for the step definition executable to still be running when cucumber is finished. Keeping the port in use, if timing is working against us just enough then this can result in an inability to open a new listening socket on that port for the BOOST test case below.

src/connectors/wire/WireServer.cpp
```diff
@@ -18,6 +18,17 @@ void SocketServer::listen(const port_type port) {
     acceptor.listen(1);
 }
 
+SocketServer::port_type SocketServer::listenPort() const {
+  boost::system::error_code ec;
+  const tcp::endpoint ep(acceptor.local_endpoint(ec));
+  if (ec)
+  {
+    return 0;
+  }
```
paoloambrosio: I don't think any part of the software can handle a failure here (return value of 0). Isn't it better to use the other signature that throws boost::system::system_error to make it crash if for some reason the port number cannot be retrieved?

If you agree, remember to change the documentation for the interface as well :-)

P.S. Position of `{` not consistent with the rest of the code

muggenhor: @paoloambrosio: ack, it's in with the new push

paoloambrosio:
One of the Travis builds failed but there is nothing wrong with this PR. The build should go back to green when the next PR is merged.

PR #133, opened 2016-12-30
Travis: output CTest error log on failures

muggenhor:
This causes CTest to output the stdout/stderr log of failing test cases when encountered.

### DIFF
--- a/travis.sh
+++ b/travis.sh
@@ -3,6 +3,9 @@ set -e #break script on non-zero exitcode from any command
 gem install bundler
 bundle install
 
+CTEST_OUTPUT_ON_FAILURE=ON
+export CTEST_OUTPUT_ON_FAILURE
+
 cmake -E make_directory build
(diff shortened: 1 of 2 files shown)


travis.sh
```diff
@@ -3,6 +3,9 @@ set -e #break script on non-zero exitcode from any command
 gem install bundler
 bundle install
 
+CTEST_OUTPUT_ON_FAILURE=ON
+export CTEST_OUTPUT_ON_FAILURE
```
muggenhor: No, that's a short hand in bash and other korn shell derivatives. Bourne shell (/bin/sh) doesn't support that.

paoloambrosio: Fair enough. Wasn't aware that Dash (`/bin/sh` in most Linux distributions, as opposed to Bash in OSX) did not support that.

konserw:
I think this should be merged - @paoloambrosio  ?

muggenhor:
FYI: I think it's ready to be merged.

PR #134, opened 2017-01-07
Test with Valgrind, stop leaking, more memory safety

muggenhor:
In short this change does these things:
 * Adds running of the unit tests with valgrind which can be enabled with the CMake parameter `-DVALGRIND_TESTS=ON`
 * Add another job on Travis which uses this
 * Reduce memory allocations
 * Don't use dynamically allocated memory for global variables
   - i.e. don't store them as global pointers, instead store them globally by-value
   - this still doesn't suffer from the mentioned static-initialisation-order fiasco, because that only applies to namespace- and class-scoped variables, not function-scoped static variables
 * Remove all uses of `new` in favor of `make_shared` in `src/`, except for:
   - when we're building a `scoped_ptr` (which lacks such a factory function)
   - the log interceptor for Boost's test driver, because that test driver assumes responsibility for deallocation of the logger
 * Make ownership a lot clearer: use references where no ownership is involved if possible, smart pointers where it is

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,10 @@ matrix:
       env: GMOCK_PATH=/usr/src/gmock
     - os: osx
       compiler: gcc #does anyone on osx use it?
+  include:
+    - os: linux
+      compiler: gcc
+      env: GMOCK_VER=1.8.0 VALGRIND_TESTS=ON
 
 addons:
   apt:
@@ -30,6 +34,7 @@ addons:
       - libboost-test-dev
       - google-mock
       - ninja-build
+      - valgrind
 
 before_install:
   - if [[ "${TRAVIS_OS_NAME}" = "osx" ]]; then brew update && brew install ninja; fi
--- a/travis.sh
+++ b/travis.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
 set -e #break script on non-zero exitcode from any command
+set -x #display command being executed
 gem install bundler
 bundle install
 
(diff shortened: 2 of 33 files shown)


.travis.yml
```diff
@@ -39,7 +44,3 @@ script: ./travis.sh
 notifications:
   email:
     - cukes-devs@googlegroups.com
-
-branches:
-  only:
-    - master
```
paoloambrosio: We might want to do this, but not as part of this PR. Guess it's a leftover.

travis.sh
```diff
@@ -1,12 +1,17 @@
 #!/bin/sh
 set -e #break script on non-zero exitcode from any command
+set -x #display command being executed
```
paoloambrosio: Is this intentional?

travis.sh
```diff
@@ -1,12 +1,17 @@
...
 gem install bundler
 bundle install
 
+CTEST_OUTPUT_ON_FAILURE=ON
+export CTEST_OUTPUT_ON_FAILURE
+
```
paoloambrosio: Already part of #133

travis.sh
muggenhor: I don't remember, I can remove it if you wish.

.travis.yml
muggenhor: Ah yes, I needed that to be able to test in my own fork. I'll split it off to a separate PR

muggenhor:
> My only problem is with changing strings to char *s (c3fdcdc). IMO it's less readable and there aren't enough allocations to justify it.

I initially did this because it was the leading allocation (that didn't get deallocated). That was the result of StepInfo instances never getting free'd. As for readability, I think the only place where it's decreasing that is in the `toSourceString` function. The rest is mostly the same. But as it's not needed anymore I'll take it out.

travis.sh
paoloambrosio: Actually let's leave it there. It would make it consistent with AppVeyor and provide value :+1:

.travis.yml
paoloambrosio: I think it would be better to separate this change and implement it for AppVeyor as well. As it's a trivial change and a PR wouldn't add anything to it, I'll push the change without PR.

travis.sh
paoloambrosio: Merged now

PR #135, opened 2017-01-07
GMock: fix dependency declarations

muggenhor:
This ensures CMake knows how to build the library files, so that if they're used as dependencies it knows how to order the build chain, even on fully parallelized builds. Additionally this switches Travis over to use Ninja for a build system instead of Make, which makes use of those parallel builds.

This partially replaces what PR #109 does and unlike that PR it's mergeable with master. It will also have a merge conflict with #130, but I've been careful to do as few changes as possible to keep that conflict as small as possible. At this time I'm most interested in getting this reviewed. If it's more convenient I'm happy to rebase this after merging #130.

Logically this change does one thing only (inside `FindGMock.cmake`): declare the produced libraries as being produced by the `ExternalProject_Add` declarations. (There's also the fix of a small typo in the `libgmock` import declaration which rendered that target completely useless).


konserw:
This is exactly what I tried to achieve with #109 but I'm not as good with CMake as you :) If this is merged I'll close my PR.

muggenhor:
Yeah CMake is not the easiest to get started with. There's a lot of bad tutorials out there and CMake's own documentation is more of a reference manual than a guide.

Anyway thanks for your vote of confidence!

paoloambrosio:
I believe this assumes that Ninja requires a superset of what is required for Make (given that we are not going to compile with Make anymore). Am I correct?

Also it's unfortunate that Ninja is not enabled on AppVeyor given how slow NMake is (I believe [this](https://github.com/boostorg/hana/blob/master/.appveyor.yml) has an example of how to install it).

We'll wait for #130 to be merged and this rebased as you said.

muggenhor:
If you would use Make with the `-j$N` option with `$N > 1` it would require the same level of correctness from the dependencies as Ninja does. It's just that Ninja by default requires a fully correct dependency graph (it's more strict and enforces more of its requirements _before_ even starting a build) and by default already runs in multiple jobs (I think the default is `$amount_of_processors + 1`).

I've got experience with using Ninja on Linux and OSX (OSX only through CI systems at work and Travis), I have none for Windows. So I think it's smart to perform experiments for using it on Windows separately from this PR.

paoloambrosio:
#130 merged now :)

muggenhor:
Hmm, apparently `BUILD_BYPRODUCTS` only gets `<BINARY_DIR>` expanded on CMake 3.3+. Will rewrite that to the manual approach.

muggenhor:
If this builds it can be merged (provided no one has review comments).

paoloambrosio:
Tried it with "make -j5" and it failed (see gist https://gist.github.com/paoloambrosio/787377fe5229e286bb742bafcefb4e68) but with "make" it works. GNU Make 4.1 on Ubuntu 16.04

Any idea why?

muggenhor:
Could you show the content of `/home/blues/Workspace/Cucumber-CPP/cucumber-cpp-135-ninja/build/gmock/src/gmock-stamp/gmock-configure-*.log`? That should show why configuring the external project fails.

paoloambrosio:
I can do that later, but does it really work for you with "-j5"? It fails even when I run it on a completely different machine with OSX and GNU Make 3.81.

paoloambrosio:
I will do it later when I have access to that machine. The same thing happens on OSX with GNU Make 3.81.

```
mkdir build && cd build
cmake ..
make -j5
```

Can you confirm that `make -j5` works for you in a fresh build directory (running parallel make for the first time). Either I'm doing something completely stupid or it must be repeatable!

muggenhor:
My full set of commands, that work with c67dbe4:
```sh
rm -rf build
mkdir build
cmake -E chdir build cmake \
    -DCUKE_ENABLE_EXAMPLES=ON \
    -DCUKE_DISABLE_BOOST_TEST=OFF \
    -DGMOCK_VER=$x \
    -G Unix\ Makefiles ..
make -C build -j5
```

Where x=1.6.0, x=1.7.0 and x=1.8.0.

I am running this on a Debian Stretch laptop with only two cores though, so the diminished concurrency may be why this doesn't trigger for me.

paoloambrosio:
On the Mac I have an Intel [i5 5287U](http://ark.intel.com/products/84988/Intel-Core-i5-5287U-Processor-3M-Cache-up-to-3_30-GHz) that has 2 cores but 4 threads, on Linux an Intel [i7 3720QM](http://ark.intel.com/products/64891/Intel-Core-i7-3720QM-Processor-6M-Cache-up-to-3_60-GHz?q=3720QM) with 4 cores and 8 threads. Both have an SSD.

Unfortunately Travis is too slow and does not show the issue.

paoloambrosio:
I've updated the gist, but there isn't much information in those logs.

muggenhor:
Hmm, just a warning, that's indeed not much.

What CMake version are you using? I'm running 3.7.1, maybe an older version contains some bug or something (although they generally try to be backwards compatible even with bugs...).

muggenhor:
Okay this is strange. I do suspect a bug in CMake now, because for Ninja it properly generates a build rule for `libgtest.a`. The makefile generator does no such thing however. This can be circumvented with an explicit `add_dependencies` though, so I've added that.

Do you still see this problem now?

paoloambrosio:
I've tried with Ninja as well and it doesn't work either. Not sure if you pushed changes.

From the Ninja output it is clear that it tries to compile GTestDriver.cpp before it has downloaded GoogleTest:
```
[15/85] Building CXX object src/CMakeFiles/cucumber-cpp.dir/drivers/GTestDriver.cpp.o
FAILED: src/CMakeFiles/cucumber-cpp.dir/drivers/GTestDriver.cpp.o 
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++    -I/Users/blues/Workspace/Cucumber-CPP/deps/include -I../include -Igmock/src/gtest/include -DBOOST_ALL_DYN_LINK -MD -MT src/CMakeFiles/cucumber-cpp.dir/drivers/GTestDriver.cpp.o -MF src/CMakeFiles/cucumber-cpp.dir/drivers/GTestDriver.cpp.o.d -o src/CMakeFiles/cucumber-cpp.dir/drivers/GTestDriver.cpp.o -c /Users/blues/Workspace/Cucumber-CPP/cucumber-cpp-135-ninja/src/drivers/GTestDriver.cpp
/Users/blues/Workspace/Cucumber-CPP/cucumber-cpp-135-ninja/src/drivers/GTestDriver.cpp:3:10: fatal error: 'gtest/gtest.h' file not found
#include <gtest/gtest.h>
         ^
1 error generated.
[18/85] Performing download step (git clone) for 'gmock'
-- gmock download command succeeded.  See also /Users/blues/Workspace/Cucumber-CPP/cucumber-cpp-135-ninja/build/gmock/src/gmock-stamp/gmock-download-*.log
[20/85] Performing download step (git clone) for 'gtest'
-- gtest download command succeeded.  See also /Users/blues/Workspace/Cucumber-CPP/cucumber-cpp-135-ninja/build/gmock/src/gtest-stamp/gtest-download-*.log
ninja: build stopped: subcommand failed.
```

Versions: CMake 3.7.2 and Ninja 1.7.2.

muggenhor:
The problem you see is definitely caused by a missing dependency of the `cucumber-cpp` library targets on `GTest::GTest`. I've pushed another commit that adds that change, but I'm not sure that's enough, looking at the generated Makefiles I still cannot find the dependency I'm now expecting to see.

muggenhor:
Okay, so apparently that's not improving things. I'm a bit rushed right now, busy with work stuff, so it may be that I'll only be able to look at this again on Monday.

konserw:
I have had similar issue and resolved it by adding:
```
if(GMOCK_FOUND) 
    add_dependencies(cucumber-cpp gmock) 
    add_dependencies(cucumber-cpp-nomain gmock) 
endif()
```
at the end of `cucumber-cpp/src/CMakeLists.txt`.
It is not the most elegant solution though.

muggenhor:
Right, so these kind of `add_dependencies` (which don't affect usage) do fix things for the Ninja generator:
`add_dependencies(GTest::GTest gtest)`

Somehow the Makefile generator still doesn't produce the correct dependency though. This seems to be related to this sentence from CMake's documentation on `add_dependencies`:
> Dependencies added to an imported target or an interface library are followed transitively in its place since the target itself does not build.

That's relevant because we're dealing with an imported target. Except that it's not clear what they mean: adding a dependency on an imported target or a dependency of an imported target on something else.

@konserw: your approach produces no differences at all in the produced makefiles for me. Unlike these `add_dependencies(GTest::GTest gtest)` which do cause several additional dependencies to be emitted by the Makefile generator.

For reference this is the diff between not having the `add_dependencies` calls in `FindGMock.cmake` and having them:
```diff
diff -Nur build.orig/CMakeFiles/Makefile2 build/CMakeFiles/Makefile2
--- build.orig/CMakeFiles/Makefile2	2017-01-30 14:56:22.021991467 +0100
+++ build/CMakeFiles/Makefile2	2017-01-30 15:00:54.778795893 +0100
@@ -132,6 +132,7 @@
 # Target rules for target CMakeFiles/e2e-steps.dir
 
 # All Build rule for target.
+CMakeFiles/e2e-steps.dir/all: CMakeFiles/gmock.dir/all
 CMakeFiles/e2e-steps.dir/all: src/CMakeFiles/cucumber-cpp.dir/all
 	$(MAKE) -f CMakeFiles/e2e-steps.dir/build.make CMakeFiles/e2e-steps.dir/depend
 	$(MAKE) -f CMakeFiles/e2e-steps.dir/build.make CMakeFiles/e2e-steps.dir/build
@@ -248,7 +249,7 @@
 # Target rules for target src/CMakeFiles/cucumber-cpp-nomain.dir
 
 # All Build rule for target.
-src/CMakeFiles/cucumber-cpp-nomain.dir/all:
+src/CMakeFiles/cucumber-cpp-nomain.dir/all: CMakeFiles/gmock.dir/all
 	$(MAKE) -f src/CMakeFiles/cucumber-cpp-nomain.dir/build.make src/CMakeFiles/cucumber-cpp-nomain.dir/depend
 	$(MAKE) -f src/CMakeFiles/cucumber-cpp-nomain.dir/build.make src/CMakeFiles/cucumber-cpp-nomain.dir/build
 	@$(CMAKE_COMMAND) -E cmake_echo_color --switch=$(COLOR) --progress-dir=/home/giel/git/cucumber-cpp/build/CMakeFiles --progress-num=53,54,55,56,57,58,59,60,61,62,63,64,65,66,67,68,69 "Built target cucumber-cpp-nomain"
```
Full diff here, but it's just more of the same: http://pastebin.com/YxtqCSz1

As I see it exactly those dependencies (of `cucumber-cpp-nomain.dir/all` on `gmock.dir/all`) should be enough to get this working correctly, even in parallel builds. Although looking at the output produced by Ninja, those are even stronger, where it lists a dependency on the `gmock` target for every object file to be built for `cucumber-cpp-nomain`:
```ninja
build cmake_order_depends_target_cucumber-cpp-nomain: phony || gmock
build src/CMakeFiles/cucumber-cpp-nomain.dir/drivers/GTestDriver.cpp.o: CXX_COMPILER__cucumber-cpp-nomain ../src/drivers/GTestDriver.cpp || cmake_order_depends_target_cucumber-cpp-nomain
```

So maybe that's necessary there: the object file dependencies, in which case we're possibly dealing with a bug in CMake's Makefile generator.

muggenhor:
I downloaded the generated `build.ninja` from Travis (after basically catting it in the log output) and see plenty of differences, but the same `cmake_order_depends_target_XXX: gmock` and `XXX/drivers/GTestDriver.cpp.o: cmake_order_depends_target_XXX` rules exist for Travis as well. That `gmock` target includes `gmock-download` which is only done when `git clone` succeeded. So I'm currently stumped why this is happening (also I'm wondering why I cannot reproduce this on any of my own systems).

konserw:
I haven't tried this before, but now (at  	8a9a61a ) I'm able to compile using both `ninja` and `make -j5` on my linux host - so you've probably nailed it 👍

muggenhor:
Had to fix a remaining merge problem. (Use of `libgtest/libgmock` name instead of `GTest/GMock`.) It should now be ready to go. Provided that @paoloambrosio cannot reproduce the build problem anymore.

paoloambrosio:
It still fails, and now I've created a Docker container to run those builds to verify that it's not my project setup: https://github.com/paoloambrosio/cucumber-cpp-build

Now this PR doesn't solve the problem of parallel builds in general (it still doesn't work for me), but reduces the likelihood of failures (it works for you two). We are stalled as you cannot reproduce the issue and I don't have time to work on it. I suggest we merge this PR anyway and keep working on other issues. What are your thoughts?

muggenhor:
@paoloambrosio I think that may be the best approach.

muggenhor:
After testing a bit more with your docker container the problem seems to be that the `git clone` command returns _before_ it finished checking out. I.e. if I change `build/gmock/tmp/gmock-gitclone.cmake` (after running CMake, before running make/ninja) and add this it sometimes shows an empty directory (except for a .git dir) and sometimes shows a filled one:
```cmake
execute_process(COMMAND sh -c "ls -lA /usr/src/cucumber-cpp/build/gmock/src/gtest > /dev/tty")
```

If I change the `git clone` command to sleep 0.1 seconds it happens less frequently but still happens. At 1 second it just doesn't happen.

paoloambrosio:
YES! YES! YES! GREAT JOB!

muggenhor:
Thanks for your dockerfile and readme. I'm not sure I would have been able to reproduce and fix this problem without the accompanying readme.

PR #136, opened 2017-01-07
Remove email notifications from CI to the Google Group

paoloambrosio:
Removed email notification from the CI build to stop flooding the [cukes-devs](https://groups.google.com/forum/#!forum/cukes-devs) group with dozens of build messages from private repos.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,10 +36,8 @@ before_install:
 
 script: ./travis.sh 
 
-notifications:
-  email:
-    - cukes-devs@googlegroups.com
-
 branches:
   only:
     - master
+
+# default notifications: https://docs.travis-ci.com/user/notifications#Notifications
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -62,7 +62,7 @@ branches:
 notifications:
 - provider: Email
   to:
-  - cukes-devs@googlegroups.com
+  - '{{commitAuthorEmail}}'
   on_build_success: false
   on_build_failure: true
   on_build_status_changed: true


.travis.yml
```diff
@@ -33,8 +33,7 @@ addons:
 script: ./travis.sh 
 
 notifications:
-  email:
-    - cukes-devs@googlegroups.com
+  email: false
```
muggenhor: Why not just remove this section from the config file and use the defaults instead?

[The default build notifications](https://docs.travis-ci.com/user/notifications#Notifications):
>  default, email notifications are sent to the committer and the commit author, if they are members of the repository (that is, they have push or admin permissions for public repositories, or if they have pull, push or admin permissions for private repositories).
> 
> Emails are sent when, on the given branch:
> 
>   * a build was just broken or still is broken
>   * a previously broken build was just fixed

With the change to `false` you're completely disabling notifications. I find them rather useful the way they're set by default.

paoloambrosio: Can we achieve the same alerting on AppVeyor?

muggenhor: For the part to only trigger on a change from failure to success (instead of every success) I think so: https://www.appveyor.com/docs/notifications/#triggering-notifications

But for sending mails to the commit author and committer I cannot find anything in the docs there.

paoloambrosio: I meant sending mails to the commit author and committer (that I agree is useful).

I'd rather have both CI configured in the same way, or committers would have to check in two places (or more likely ignore Windows failures).

muggenhor: I share the desire for both CI systems to behave the same there, but I cannot see how to accomplish that from the AppVeyor docs. So I'd rather have at least one of them (Travis) be as useful as possible instead of both be equally uncommunicative (i.e. behaving no better than the lowest common demoninator IMO translates to "perfect is the enemy of good").

Maybe it is possible with AppVeyor, but it is not (easily) findable in the documentation.

paoloambrosio:
@muggenhor apparently now there is a documented way in AppVeyor to send notifications to the commit author: https://www.appveyor.com/docs/notifications/#appveyoryml-configuration

I've tested it with another branch and it seems to work. Happy with this?

muggenhor:
@paoloambrosio this looks good to me

Issue #119, opened 2016-09-03
clang >= 3.7 incompatibility


konserw:
Hello,
I've just noticed 2 problems when compiling with clang version 3.8.1 on GNU/Linux (most probably problem exists for clang >= 3.7 both on Linux and  MAC) 
- WireProtocols test are generating such warning:
  `expression with side effects will be evaluated despite being used as an operand to 'typeid'`
  on lines where `EXPECT_PTRTYPE` macro is used
- SocketServer tests are failing with some exception: 
  `The following tests FAILED:
   21 - SocketServerTest.exitsOnFirstConnectionClosed (OTHER_FAULT)
   22 - SocketServerTest.moreThanOneClientCanConnect (OTHER_FAULT)
   23 - SocketServerTest.receiveAndSendsSingleLineMassages (OTHER_FAULT)`

Those 2 issues may or may not be connected with each other. Everything works perfectly fine using gcc 6.1.1 and of course on our CI builds.

I'm not sure how to fix that, but please let me know if I can help with this ;)

konserw:
Socketserver tests appear to be fixed by #120 - not sure how. But warnings are still to be investigated in my opinion.
BTW. I'm very curious how #120 fixed those tests ;)

paoloambrosio:
@konserw what Linux distro are you using? I can't reproduce it on Ubuntu with clang++ 3.8.0-2ubuntu4

konserw:
Arch Linux (Antegros)
clang --version
clang version 3.9.1 (tags/RELEASE_391/final)
Target: x86_64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin

But the warning is also visible on travis OSX jobs, for instance: https://travis-ci.org/cucumber/cucumber-cpp/jobs/214419938

paoloambrosio:
I must have done something wrong the first time. Now I can see the warning :-)

PR #138, opened 2017-03-23
Get rid of clang warning - fix for issue 119

konserw:
This change removes CLang warning about typeid on pointer, so fixes issue #119

### DIFF
--- a/tests/integration/WireProtocolTest.cpp
+++ b/tests/integration/WireProtocolTest.cpp
@@ -21,9 +21,6 @@ public:
     MOCK_CONST_METHOD3(snippetText, std::string(const std::string & keyword, const std::string & name, const std::string & multilineArgClass));
 };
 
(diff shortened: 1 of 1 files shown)


tests/integration/WireProtocolTest.cpp
```diff
@@ -25,9 +25,7 @@ class MockCukeEngine : public CukeEngine {
     EXPECT_THAT(typeid(classname).name(), StrEq(typeid(expression).name()))
```
paoloambrosio: I think `EXPECT_TYPE` was only needed for `EXPECT_PTRTYPE`. Guess we can remove it now.

paoloambrosio:
Squashed it myself without thinking GitHub wouldn't be able to cross-reference this PR. Anyway the squashed commits have been merged into master with f933ad1983cf15cc0573532f98b5457bc1ba2a18.

PR #139, opened 2017-03-27
Fix and document minimal boost version

muggenhor:
Make Boost `REQUIRED` at CMake time so that we get nice error handling at that time. Require at least version 1.42 for `Boost.Test` support (because it doesn't build until that version). Also update the documentation to match the implementation.


konserw:
👍 good PR - CMake now works well with and without boost ;)

muggenhor:
I'm assuming you mean without Boost.Test ;-) Because it shouldn't allow you to configure without Boost.

konserw:
Yes - it failed gracefully without any boost :)

PR #140, opened 2017-03-30
Enable Qt framework in CI

konserw:
Main purpose of this PR is to setup Qt framework in CI on all 3 platforms:
- Linux via apt
- OSX via brew
- Windows using preinstalled Qt in appveyor (only mingwx64 is missing on the image)
Rework of Qt-related code in CMake files was needed to prepare for QtTest driver and ensure that CalcQt can be build under all Qt versions.
I've also done some refactoring in appveyor configuration: 
- removed not used variables
- moved  build steps to batch file
- ruby version is now hard coded instead of variable
I've also changed how travis matrix is described in config but still build targets are the same. I hope now config is more clear.

Please leave a comment how you like those changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: cpp
 os:
-  - osx
   - linux
 sudo: required
 dist: trusty
@@ -13,14 +12,19 @@ env:
   - GMOCK_PATH=/usr/src/gmock #1.6.0 from ubuntu trusty repo
 matrix:
   exclude:
-    - os: osx
-      env: GMOCK_PATH=/usr/src/gmock
-    - os: osx
-      compiler: gcc #does anyone on osx use it?
+    - os: linux
+      compiler: gcc
+      env: GMOCK_VER=1.8.0
   include:
     - os: linux
       compiler: gcc
       env: GMOCK_VER=1.8.0 VALGRIND_TESTS=ON
+    - os: osx
+      compiler: clang
+      env: GMOCK_VER=1.8.0 QTDIR=/usr/local/Cellar/qt5/5.8.0_1
+    - os: osx
+      compiler: clang
+      env: GMOCK_VER=1.7.0 QTDIR=/usr/local/Cellar/qt5/5.8.0_1
 
 addons:
   apt:
@@ -35,9 +39,10 @@ addons:
       - google-mock
       - ninja-build
       - valgrind
+      - qtbase5-dev
 
 before_install:
-  - if [[ "${TRAVIS_OS_NAME}" = "osx" ]]; then brew update && brew install ninja; fi
+  - if [[ "${TRAVIS_OS_NAME}" = "osx" ]]; then brew update && brew install ninja && brew install qt5; fi
 
 script: ./travis.sh 
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,9 @@ set(CUKE_ENABLE_EXAMPLES OFF CACHE BOOL "Enable the examples")
 set(GMOCK_SRC_DIR "" CACHE STRING "Google Mock framework sources path (otherwise downloaded)")
 set(GMOCK_VER "1.7.0" CACHE STRING "Google Mock framework version to be used")
 option(VALGRIND_TESTS "Run tests within Valgrind" OFF)
+set(CUKE_DISABLE_QT OFF CACHE BOOL "Disable using Qt framework")
 
+set(ignoreMe "${QT_QMAKE_EXECUTABLE}") #supress warning
 set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
 
 #
@@ -104,6 +106,34 @@ if(NOT CUKE_DISABLE_GTEST)
 endif()
 
 #
+# Qt
+#
+
+if(NOT CUKE_DISABLE_QT)
+    set(CMAKE_PREFIX_PATH $ENV{QTDIR})
+    find_package(Qt5Core  QUIET)
+    find_package(Qt5Widgets  QUIET)
+    find_package(Qt5Test  QUIET)
+
+    if(${Qt5Core_FOUND} AND ${Qt5Widgets_FOUND} AND ${Qt5Test_FOUND})
+        message(STATUS "Found Qt version: ${Qt5Core_VERSION_STRING}")
+        include_directories(${Qt5Core_INCLUDE_DIRS} ${Qt5Widgets_INCLUDE_DIRS} ${Qt5Test_INCLUDE_DIRS})
+#        set(CMAKE_INCLUDE_CURRENT_DIR ON)
+#        set(CMAKE_AUTOMOC ON)
+        if (Qt5_POSITION_INDEPENDENT_CODE)
+            SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
+        endif()
+        set(QT_LIBRARIES Qt5::Core Qt5::Widgets Qt5::Test)
+    else()
+        find_package(Qt4 COMPONENTS QtCore QtGui QtTest)
+        if(QT4_FOUND)
+            set(QT_LIBRARIES Qt4::QtCore Qt4::QtGui Qt4::QtTest)
+            include(${QT_USE_FILE})
+        endif()
+    endif()
+endif()
+
+#
 # Valgrind
 #
 
@@ -154,11 +184,8 @@ endif()
 #
 
 set(CUKE_INCLUDE_DIR ${CMAKE_CURRENT_SOURCE_DIR}/include)
-
 include_directories(${CUKE_INCLUDE_DIR})
-
 set(CUKE_LIBRARIES cucumber-cpp ${CUKE_EXTRA_LIBRARIES})
-
 add_subdirectory(src)
 
 #
@@ -222,4 +249,16 @@ endif()
 
 if(CUKE_ENABLE_EXAMPLES)
     add_subdirectory(examples)
+
+    #check if c++11 is nedded
+    if(${Qt5Core_FOUND})
+        find_program(QMAKE_EXECUTABLE NAMES qmake HINTS ${QTDIR} ENV QTDIR PATH_SUFFIXES bin)
+        execute_process(COMMAND ${QMAKE_EXECUTABLE} -query QT_VERSION OUTPUT_VARIABLE QT_VERSION)
+        if(QT_VERSION GREATER 5.6.99)
+            message(STATUS "C++11 is needed from Qt version 5.7.0, building with c++11 enabled")
+            set_property(TARGET libcalcqt PROPERTY CXX_STANDARD 11)
+            set_property(TARGET calcqt PROPERTY CXX_STANDARD 11)
+        endif()
+    endif()
 endif()
+
--- a/appveyor.bat
+++ b/appveyor.bat
@@ -1,0 +1,20 @@
+set PATH=C:\Ruby200\bin;%BOOST_LIBRARYDIR%;%PATH%
+if defined MINGW_ROOT set PATH=%MINGW_ROOT%\bin;C:\msys64\usr\bin\;%PATH%
+if defined QT_DIR set PATH=%QT_DIR%\bin;%PATH%
+if "%CMAKE_GENERATOR%"=="NMake Makefiles" call "%VS140COMNTOOLS%\..\..\VC\vcvarsall.bat" %PLATFORM%
+echo %PATH%
+
+git submodule init
+git submodule update
+call gem install bundle
+call bundle install
+if defined MINGW_ARCH bash -lc "pacman --needed --noconfirm -S mingw-w64-%MINGW_ARCH%-boost
+
+cmake -E make_directory build
+cmake -E chdir build cmake -G "%CMAKE_GENERATOR%" -DCUKE_ENABLE_EXAMPLES=on -DBOOST_ROOT="%BOOST_ROOT%" -DBOOST_INCLUDEDIR="%BOOST_INCLUDEDIR%" -DBOOST_LIBRARYDIR="%BOOST_LIBRARYDIR%" -DQT_QMAKE_EXECUTABLE="%QT_DIR%/bin/qmake.exe" ..
+cmake --build build
+
+set CTEST_OUTPUT_ON_FAILURE=ON
+cmake --build build --target test
+cmake --build build --target features
+
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,59 +2,35 @@ version: "{branch}-ci-{build}"
 os: Visual Studio 2015
 
 environment:
-  RUBY_VERSION: 200
   matrix:
-  - build: mingw
-    platform: x86
+  - CMAKE_GENERATOR: 'MSYS Makefiles'
     MINGW_ARCH: i686
-    MSYSTEM: MINGW32
     MINGW_ROOT: C:\msys64\mingw32
     BOOST_ROOT: C:\msys64\mingw32
     BOOST_LIBRARYDIR: C:\msys64\mingw32\lib
     BOOST_INCLUDEDIR: C:\msys64\mingw32\include\boost
-    CMAKE_GENERATOR: 'MSYS Makefiles'
-  - build: mingw
-    platform: x64
+    QT_DIR: C:\Qt\5.8\mingw53_32
+  - CMAKE_GENERATOR: 'MSYS Makefiles'
     MINGW_ARCH: x86_64
-    MSYSTEM: MINGW64
     MINGW_ROOT: C:\msys64\mingw64
     BOOST_ROOT: C:\msys64\mingw64
     BOOST_LIBRARYDIR: C:\msys64\mingw64\lib
     BOOST_INCLUDEDIR: C:\msys64\mingw64\include\boost
-    CMAKE_GENERATOR: 'MSYS Makefiles'
-  - build: msvc
+  - CMAKE_GENERATOR: 'NMake Makefiles'
     platform: x86
     BOOST_ROOT: C:\Libraries\boost_1_59_0
     BOOST_LIBRARYDIR: C:\Libraries\boost_1_59_0\lib32-msvc-14.0
     BOOST_INCLUDEDIR: C:\Libraries\boost_1_59_0\boost
-    CMAKE_GENERATOR: 'NMake Makefiles'
-  - build: msvc
+    QT_DIR: C:\Qt\5.8\msvc2015
+  - CMAKE_GENERATOR: 'NMake Makefiles'
     platform: x64
     BOOST_ROOT: C:\Libraries\boost_1_59_0
     BOOST_INCLUDEDIR: C:\Libraries\boost_1_59_0\boost
     BOOST_LIBRARYDIR: C:\Libraries\boost_1_59_0\lib64-msvc-14.0
-    CMAKE_GENERATOR: 'NMake Makefiles'
-
-install:
-- git submodule init
-- git submodule update
-- set PATH=C:\Ruby%RUBY_VERSION%\bin;%BOOST_LIBRARYDIR%;%PATH%
-- gem install bundle
-- bundle install
-- bundle env
-- if "%build%"=="mingw" set PATH=%MINGW_ROOT%\bin;C:\msys64\usr\bin\;%PATH%
-- if "%build%"=="mingw" bash -lc "pacman --needed --noconfirm -S mingw-w64-%MINGW_ARCH%-boost
+    QT_DIR: C:\Qt\5.8\msvc2015_64
 
 build_script:
-- cmd: if "%build%"=="msvc" call "%VS140COMNTOOLS%\..\..\VC\vcvarsall.bat" %PLATFORM%
-- cmd: cmake -E make_directory build
-- cmd: cmake -E chdir build cmake -G "%CMAKE_GENERATOR%" -DCUKE_ENABLE_EXAMPLES=ON -DBOOST_ROOT="%BOOST_ROOT%" -DBOOST_INCLUDEDIR="%BOOST_INCLUDEDIR%" -DBOOST_LIBRARYDIR="%BOOST_LIBRARYDIR%" ..
-- cmd: cmake --build build
-
-test_script:
-- cmd: set CTEST_OUTPUT_ON_FAILURE=ON
-- cmd: cmake --build build --target test
-- cmd: cmake --build build --target features
+- cmd: call appveyor.bat
 
 notifications:
 - provider: Email
--- a/examples/CalcQt/CMakeLists.txt
+++ b/examples/CalcQt/CMakeLists.txt
@@ -1,36 +1,19 @@
 project(CalcQt)
 
 set(CALCQT_HEADERS src/CalculatorWidget.h)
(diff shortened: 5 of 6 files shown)


CMakeLists.txt
```diff
@@ -13,7 +13,9 @@ set(CUKE_ENABLE_EXAMPLES OFF CACHE BOOL "Enable the examples")
 set(GMOCK_SRC_DIR "" CACHE STRING "Google Mock framework sources path (otherwise downloaded)")
 set(GMOCK_VER "1.7.0" CACHE STRING "Google Mock framework version to be used")
 option(VALGRIND_TESTS "Run tests within Valgrind" OFF)
+set(CUKE_DISABLE_QT OFF CACHE BOOL "Disable using Qt framework")
```
muggenhor: For boolean parameters I recommend using `option` instead (it's intended to give user options and only supports booleans).

CMakeLists.txt
```diff
@@ -13,7 +13,9 @@ set(CUKE_ENABLE_EXAMPLES OFF CACHE BOOL "Enable the examples")
...
 
+set(ignoreMe "${QT_QMAKE_EXECUTABLE}") #supress warning
```
muggenhor: What warning are you suppressing?

CMakeLists.txt
```diff
@@ -104,6 +106,34 @@ if(NOT CUKE_DISABLE_GTEST)
...
+
+if(NOT CUKE_DISABLE_QT)
+    set(CMAKE_PREFIX_PATH $ENV{QTDIR})
+    find_package(Qt5Core  QUIET)
+    find_package(Qt5Widgets  QUIET)
+    find_package(Qt5Test  QUIET)
+
+    if(${Qt5Core_FOUND} AND ${Qt5Widgets_FOUND} AND ${Qt5Test_FOUND})
+        message(STATUS "Found Qt version: ${Qt5Core_VERSION_STRING}")
+        include_directories(${Qt5Core_INCLUDE_DIRS} ${Qt5Widgets_INCLUDE_DIRS} ${Qt5Test_INCLUDE_DIRS})
```
muggenhor: Please don't add these to the global search path. Instead use `target_include_directories(${target} <PRIVATE|PUBLIC> ${dirs})` for the targets that need it.

Or better yet, only use `target_link_libraries(${target} <PRIVATE|PUBLIC> Qt5::${required_component})` because that already sets up the include path correctly.

.travis.yml
```diff
@@ -35,9 +39,10 @@ addons:
       - google-mock
       - ninja-build
       - valgrind
+      - qtbase5-dev
 
 before_install:
-  - if [[ "${TRAVIS_OS_NAME}" = "osx" ]]; then brew update && brew install ninja; fi
+  - if [[ "${TRAVIS_OS_NAME}" = "osx" ]]; then brew update && brew install ninja && brew install qt5; fi
```
muggenhor: Question: doesn't brew support installing multiple packages at once? E.g. `brew install ninja qt5` ?

CMakeLists.txt
```diff
@@ -104,6 +106,34 @@ if(NOT CUKE_DISABLE_GTEST)
 endif()
 
 #
+# Qt
+#
+
+if(NOT CUKE_DISABLE_QT)
+    set(CMAKE_PREFIX_PATH $ENV{QTDIR})
+    find_package(Qt5Core  QUIET)
```
muggenhor: Why do you search `QUIET` for these packages? I think it's rather helpful to have the output of these searches shown on the console.

CMakeLists.txt
```diff
@@ -222,4 +249,16 @@ endif()
 
 if(CUKE_ENABLE_EXAMPLES)
     add_subdirectory(examples)
+
+    #check if c++11 is nedded
+    if(${Qt5Core_FOUND})
+        find_program(QMAKE_EXECUTABLE NAMES qmake HINTS ${QTDIR} ENV QTDIR PATH_SUFFIXES bin)
+        execute_process(COMMAND ${QMAKE_EXECUTABLE} -query QT_VERSION OUTPUT_VARIABLE QT_VERSION)
+        if(QT_VERSION GREATER 5.6.99)
```
muggenhor: Please use this instead:
```cmake
if(NOT QT_VERSION VERSION_LESS 5.7)
```

That does a proper lexicographic comparison of the version number.

CMakeLists.txt
```diff
@@ -222,4 +249,16 @@ endif()
 
 if(CUKE_ENABLE_EXAMPLES)
     add_subdirectory(examples)
+
+    #check if c++11 is nedded
```
muggenhor: Typo: "needed"

CMakeLists.txt
```diff
@@ -222,4 +249,16 @@ endif()
...
+    if(${Qt5Core_FOUND})
+        find_program(QMAKE_EXECUTABLE NAMES qmake HINTS ${QTDIR} ENV QTDIR PATH_SUFFIXES bin)
+        execute_process(COMMAND ${QMAKE_EXECUTABLE} -query QT_VERSION OUTPUT_VARIABLE QT_VERSION)
```
muggenhor: Can't you just use `Qt5Core_VERSION_STRING` instead?

CMakeLists.txt
konserw: Something like Not used variable passed to CMake, and this variable is needed for appveyor CI

CMakeLists.txt
konserw: I'll change it

CMakeLists.txt
konserw: I wasn't sure how to do it, becouse if finding qt5 failed i'm searching for qt4 which is sufficient for the examples. REQUIRED would fail the configuration. Mabey I'll go without quiet

CMakeLists.txt
konserw: I'll check

CMakeLists.txt
muggenhor: You don't have to specify either one of those. I.e. just this should work:
```cmake
find_package(Qt5Core)
```

CMakeLists.txt
muggenhor: I don't think it's a good idea to silence that warning. Because when it gives that warning the parameter you passed on the command line actually isn't used. If that's okay under some circumstances then I would just allow the warning to be displayed.

CMakeLists.txt
konserw: In fact I've changed all BOOLs to options - how do you feel about that?

CMakeLists.txt
konserw: done :) Didn't know that link_libraries also setup includes

CMakeLists.txt
konserw: Done - one more thing I've learned today about cmake.

.travis.yml
konserw: Done

konserw:
I've also updated CMake minimium version as @muggenhor in #142  seems to agree  that it is needed

CMakeLists.txt
muggenhor: Sounds good :+1:

CMakeLists.txt
muggenhor: It propagates all properties that have INTERFACE in the name (and when the linked library is a STATIC library also `LINK_LIBRARIES`). This is nice because it allows you to specify all usage requirements of a target and have CMake do all the work of ensuring those are met in other targets that use that one.

muggenhor:
@konserw:
> I've also updated CMake minimium version as @muggenhor in #142 seems to agree that it is needed

Ack.

3.0+ is needed to have `find_dependency()`, that is used by `FindGMock.cmake`. 3.1+ is needed to have the `Threads` package create the `Threads::Threads` `IMPORT` target.

konserw:
Should I change anything else?

appveyor.bat
```diff
@@ -0,0 +1,20 @@
+set PATH=C:\Ruby200\bin;%BOOST_LIBRARYDIR%;%PATH%
+if defined MINGW_ROOT set PATH=%MINGW_ROOT%\bin;C:\msys64\usr\bin\;%PATH%
+if defined QT_DIR set PATH=%QT_DIR%\bin;%PATH%
```
muggenhor: I think it might make more sense to add `%QT_DIR%` to [`CMAKE_PREFIX_PATH`](https://cmake.org/cmake/help/v3.5/variable/CMAKE_PREFIX_PATH.html). That should also remove the need for manually specifying `QT_QMAKE_EXECUTABLE`.

examples/CalcQt/CMakeLists.txt
```diff
@@ -1,36 +1,20 @@
...
-    set(QT_LIBRARIES Qt5::Core Qt5::Widgets Qt5::Test)
-
-    add_library(libcalcqt src/CalculatorWidget.cpp ${CALCQT_HEADERS})
+if(QT_LIBRARIES)
+    if(QT4_FOUND)
+        qt4_wrap_cpp(CALCQT_MOC ${CALCQT_HEADERS})
+    else()
+        qt5_wrap_cpp(CALCQT_MOC ${CALCQT_HEADERS})
+    endif()
+    add_library(libcalcqt src/CalculatorWidget.cpp ${CALCQT_MOC})
```
muggenhor: I think you can just use [`AUTOMOC`](https://cmake.org/cmake/help/v3.1/prop_tgt/AUTOMOC.html) here and get rid of the `qtX_wrap_cpp` stuff altogether.

CMakeLists.txt
muggenhor: See my comment about adding `%QT_DIR%` to `CMAKE_PREFIX_PATH` in `appveyor.bat`: I don't think you need to set this variable at all. So please remove this line.

muggenhor:
@konserw have you looked at my questions?

.travis.yml
```diff
@@ -13,14 +12,19 @@ env:
   - GMOCK_PATH=/usr/src/gmock #1.6.0 from ubuntu trusty repo
 matrix:
   exclude:
-    - os: osx
-      env: GMOCK_PATH=/usr/src/gmock
-    - os: osx
-      compiler: gcc #does anyone on osx use it?
+    - os: linux
+      compiler: gcc
+      env: GMOCK_VER=1.8.0
```
konserw: I thought it wasn't necessary as gmock 1.8 is used for the valgrind build. Reverted it now.

CMakeLists.txt
```diff
@@ -104,6 +105,31 @@ if(NOT CUKE_DISABLE_GTEST)
 endif()
 
 #
+# Qt
+#
+
+if(NOT CUKE_DISABLE_QT)
+    set(CMAKE_PREFIX_PATH $ENV{QTDIR})
```
konserw: I'll try, but I'm not sure how it should be done - via -D switch?

CMakeLists.txt
```diff
@@ -104,6 +105,31 @@ if(NOT CUKE_DISABLE_GTEST)
...
+    find_package(Qt5Core)
+    find_package(Qt5Widgets)
+    find_package(Qt5Test)
+
+    if(${Qt5Core_FOUND} AND ${Qt5Widgets_FOUND} AND ${Qt5Test_FOUND})
+        message(STATUS "Found Qt version: ${Qt5Core_VERSION_STRING}")
+        if (Qt5_POSITION_INDEPENDENT_CODE)
+            SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
konserw: Sorry, it is indeed not needed for this PR. PIC will be required by Qt test driver, i.e. another PR. I'll revert it here.

CMakeLists.txt
```diff
@@ -222,4 +245,16 @@ endif()
 
 if(CUKE_ENABLE_EXAMPLES)
     add_subdirectory(examples)
+
+    #check if c++11 is needed
+    if(${Qt5Core_FOUND})
+#        find_program(QMAKE_EXECUTABLE NAMES qmake HINTS ${QTDIR} ENV QTDIR PATH_SUFFIXES bin)
+#        execute_process(COMMAND ${QMAKE_EXECUTABLE} -query QT_VERSION OUTPUT_VARIABLE QT_VERSION)
```
konserw: Ok - done

CMakeLists.txt
```diff
@@ -222,4 +245,16 @@ endif()
...
+        if(NOT ${Qt5Core_VERSION_STRING} VERSION_LESS 5.7)
+            message(STATUS "C++11 is needed from Qt version 5.7.0, building with c++11 enabled")
+            set_property(TARGET libcalcqt PROPERTY CXX_STANDARD 11)
+            set_property(TARGET calcqt PROPERTY CXX_STANDARD 11)
```
konserw: Ok - didn't know that

appveyor.bat
konserw: Ok, I'll check it

examples/CalcQt/CMakeLists.txt
konserw: I thought you would like such self-contined implementation better. Automoc is fine by me :)

konserw:
@muggenhor - to be honest I just forgot about all that. Now I have just pushed changes to address outlined issues. I hope CI will be green 🤞

konserw:
It was all green so I've rebased against master

CMakeLists.txt
muggenhor: Yes, the approach you've taken in travis.sh now what I meant.

appveyor.yml
```diff
@@ -2,59 +2,35 @@ version: "{branch}-ci-{build}"
...
     BOOST_LIBRARYDIR: C:\Libraries\boost_1_59_0\lib64-msvc-14.0
-    CMAKE_GENERATOR: 'NMake Makefiles'
-
-install:
-- git submodule init
-- git submodule update
-- set PATH=C:\Ruby%RUBY_VERSION%\bin;%BOOST_LIBRARYDIR%;%PATH%
-- gem install bundle
-- bundle install
-- bundle env
```
muggenhor: I don't see `bundle env` reappearing in `appveyor.bat`. Why is that?

examples/CalcQt/CMakeLists.txt
```diff
@@ -1,36 +1,15 @@
...
-find_package(Qt5Test QUIET)
-
-if(${Qt5Core_FOUND} AND ${Qt5Widgets_FOUND} AND ${Qt5Test_FOUND})
-    set(CMAKE_INCLUDE_CURRENT_DIR ON)
-    set(CMAKE_AUTOMOC ON)
-    set(QT_LIBRARIES Qt5::Core Qt5::Widgets Qt5::Test)
-
-    add_library(libcalcqt src/CalculatorWidget.cpp ${CALCQT_HEADERS})
+if(QT_LIBRARIES)
+    add_library(libcalcqt src/CalculatorWidget.cpp)
```
muggenhor: I'm expecting `src/CalculatorWidget.h` to be present here as well. Does automoc work without specifying the header file to run Qt's moc on?

Also please remove the now unused `CALCQT_HEADERS` variable.

appveyor.yml
```diff
@@ -2,59 +2,35 @@ version: "{branch}-ci-{build}"
...
-  - build: mingw
-    platform: x64
+    QT_DIR: C:\Qt\5.8\mingw53_32
+  - CMAKE_GENERATOR: 'MSYS Makefiles'
     MINGW_ARCH: x86_64
-    MSYSTEM: MINGW64
     MINGW_ROOT: C:\msys64\mingw64
     BOOST_ROOT: C:\msys64\mingw64
     BOOST_LIBRARYDIR: C:\msys64\mingw64\lib
     BOOST_INCLUDEDIR: C:\msys64\mingw64\include\boost
```
muggenhor: Shouldn't there be a `QT_DIR` variable here (for 64-bit MinGW)?

CMakeLists.txt
muggenhor: No worries, I had to discover this the hard way myself. Unfortunately CMake's documentation is more of a reference manual than a usage manual.

examples/CalcQt/CMakeLists.txt
muggenhor: thanks :-)

appveyor.bat
```diff
@@ -0,0 +1,20 @@
+set PATH=C:\Ruby200\bin;%BOOST_LIBRARYDIR%;%PATH%
+if defined MINGW_ROOT set PATH=%MINGW_ROOT%\bin;C:\msys64\usr\bin\;%PATH%
+REM if defined QT_DIR set PATH=%QT_DIR%\bin;%PATH%
+if "%CMAKE_GENERATOR%"=="NMake Makefiles" call "%VS140COMNTOOLS%\..\..\VC\vcvarsall.bat" %PLATFORM%
+echo %PATH%
```
konserw: of course

appveyor.bat
```diff
@@ -0,0 +1,20 @@
+set PATH=C:\Ruby200\bin;%BOOST_LIBRARYDIR%;%PATH%
+if defined MINGW_ROOT set PATH=%MINGW_ROOT%\bin;C:\msys64\usr\bin\;%PATH%
+REM if defined QT_DIR set PATH=%QT_DIR%\bin;%PATH%
```
konserw: Missed that one :)

appveyor.yml
konserw: Appveyor has only 32-bit mingw-compiled qt libs preinstalled. https://www.appveyor.com/docs/build-environment/#qt
AFAIK it is not possible to link VC-compiled Qt with Mingw-compiled cucumber, or use 32-bit  mingw-compiled lib.

appveyor.yml
konserw: it is as much debugging by print as echo %path%. This is what it looks like on my machine:
$ bundle env
## Environment

```
Bundler   1.15.3
Rubygems  2.6.11
Ruby      2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
Git       2.14.1
Platform  x86_64-linux
OpenSSL   OpenSSL 1.1.0e  16 Feb 2017
```

muggenhor: Okay, removing it is fine then.

appveyor.yml
muggenhor: Thanks for your explanation

examples/CalcQt/CMakeLists.txt
```diff
@@ -1,36 +1,14 @@
 project(CalcQt)
 
-set(CALCQT_HEADERS src/CalculatorWidget.h)
-set(CALCQT_SOURCES src/CalcQt.cpp src/CalculatorWidget.cpp)
-include_directories(${CUKE_INCLUDE_DIRS} src)
```
muggenhor: I like that you remove this global `include_directories` but you should probably combine it with `target_include_directories(libcalcqt PUBLIC src)` to fix the build errors you get.

examples/CalcQt/CMakeLists.txt
```diff
@@ -1,36 +1,14 @@
...
-    set(CMAKE_AUTOMOC ON)
-    set(QT_LIBRARIES Qt5::Core Qt5::Widgets Qt5::Test)
-
-    add_library(libcalcqt src/CalculatorWidget.cpp ${CALCQT_HEADERS})
+if(QT_LIBRARIES)
+    add_library(libcalcqt src/CalculatorWidget.cpp)
     target_link_libraries(libcalcqt ${QT_LIBRARIES})
     
-    add_executable(calcqt ${CALCQT_SOURCES})
+    add_executable(calcqt src/CalcQt.cpp src/CalculatorWidget.h)
```
muggenhor: I don't think you should be adding the `CalculatorWidget.h` header file to this executable: it's logically part of the `libcalcqt` library, not this.

examples/CalcQt/CMakeLists.txt
konserw: thank you for pointing that - I haven't yet used target_include_directories

examples/CalcQt/CMakeLists.txt
konserw: of course - it's my mistake when cut-pasting ;)

CMakeLists.txt
```diff
@@ -113,6 +114,30 @@ if(NOT CUKE_DISABLE_GTEST)
...
+        set(QT_LIBRARIES Qt5::Core Qt5::Widgets Qt5::Test)
+    else()
+        find_package(Qt4 COMPONENTS QtCore QtGui QtTest)
+        if(QT4_FOUND)
+            set(QT_LIBRARIES Qt4::QtCore Qt4::QtGui Qt4::QtTest)
+            include(${QT_USE_FILE})
+        endif()
+    endif()
+    if(QT_LIBRARIES)
+        set(CMAKE_AUTOMOC ON)
```
muggenhor: Unfortunately this causes it to add a ton of extra files to compile (one for every target). Can you remove this and instead add this to `examples/CalcQt/CMakeLists.txt`?
```cmake
set_target_properties(libcalcqt PROPERTIES AUTOMOC ON)
```

CMakeLists.txt
```diff
@@ -231,4 +253,16 @@ endif()
 
 if(CUKE_ENABLE_EXAMPLES)
     add_subdirectory(examples)
+
+    #check if c++11 is needed
+    if(${Qt5Core_FOUND})
+        if(NOT ${Qt5Core_VERSION_STRING} VERSION_LESS 5.7)
+            message(STATUS "C++11 is needed from Qt version 5.7.0, building with c++11 enabled")
+            set_target_properties(calcqt libcalcqt PROPERTIES
```
muggenhor: Looking at the `Qt5CoreConfigExtras.cmake` file in [Qt 5.7.1 (as present on Debian stretch)](https://packages.debian.org/stretch/qtbase5-dev) this shouldn't be necessary because of this line:
```cmake
set_property(TARGET Qt5::Core PROPERTY INTERFACE_COMPILE_FEATURES cxx_decltype)
```

That makes it a usage requirement (`INTERFACE_\w+` are properties describing usage requirements automatically applied by CMake) to always compile this target with a compiler supporting `decltype`, that means C++11 (or C++0x on some very old compilers, but if that's not enough then that entire compiler won't work anyway). So it should already automatically switch to using C++11 for every target that links to `Qt5::Core` (via `target_link_libraries`).

So please remove this entire block.

muggenhor:
One problem, since you're changes to the AppVeyor config failures are no longer fatal (search for "undefined reference"): https://ci.appveyor.com/project/paoloambrosio/cucumber-cpp-qqrt7/build/master-ci-249/job/fxe93x2pmd8ewhlc

muggenhor:
> One problem, since you're changes to the AppVeyor config failures are no longer fatal (search for "undefined reference"): https://ci.appveyor.com/project/paoloambrosio/cucumber-cpp-qqrt7/build/master-ci-249/job/fxe93x2pmd8ewhlc

I think this is caused by your move of the script to `appveyor.bat`. Additionally that has the effect of no longer showing which command is being executed. Is there any particular reason why you've moved all this out of `appveyor.yml` and into the batch file?

CMakeLists.txt
konserw: And this is why I avoided automoc at first - didn't know it can be done with target property.

CMakeLists.txt
konserw: ok

konserw:
@muggenhor - I thought it was good idea to separate configuration from build script like we have done with travis. It just turned out badly with cmd instead of bash - merged it again - I hope it'll help.

Issue #141, opened 2017-03-30
Flaky TCPSocketServer unit test


konserw:
There is some kind of problem with TCPSocketServerTest unit test. All 4 fails:
```
The following tests FAILED:
	 21 - TCPSocketServerTest.exitsOnFirstConnectionClosed (OTHER_FAULT)
	 22 - TCPSocketServerTest.moreThanOneClientCanConnect (OTHER_FAULT)
	 23 - TCPSocketServerTest.receiveAndSendsSingleLineMassages (OTHER_FAULT)
	 24 - UnixSocketServerTest.fullLifecycle (OTHER_FAULT)
```
when I'm building cucumber with c++11 (and 14) enabled via CMake:
```
set_property(TARGET cucumber-cpp-nomain PROPERTY CXX_STANDARD 11)
set_property(TARGET cucumber-cpp PROPERTY CXX_STANDARD 11)
```
Those test failed before for me as outlined in #119 and then fixed itself after merging #120 for whatever reason.
Please help me investigate this issue.

EDIT: 
full fail text:
```
21/91 Test #21: TCPSocketServerTest.exitsOnFirstConnectionClosed .................................***Exception: Other  0.07 sec
Running main() from gmock_main.cc
Note: Google Test filter = TCPSocketServerTest.exitsOnFirstConnectionClosed
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from TCPSocketServerTest
[ RUN      ] TCPSocketServerTest.exitsOnFirstConnectionClosed
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >'
  what():  accept: Already open

      Start 22: TCPSocketServerTest.moreThanOneClientCanConnect
22/91 Test #22: TCPSocketServerTest.moreThanOneClientCanConnect ..................................***Exception: Other  0.08 sec
Running main() from gmock_main.cc
Note: Google Test filter = TCPSocketServerTest.moreThanOneClientCanConnect
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from TCPSocketServerTest
[ RUN      ] TCPSocketServerTest.moreThanOneClientCanConnect
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::system::system_error> >'
  what():  accept: Already open

[...]

PR #142, opened 2017-03-30
QtTest driver v2

konserw:
QtTest based test driver for cucumber-cpp.
Includes also QtTestSteps for both Calc and CalcQt examples.
It is based of PR #140 for Qt in CI
Please take a look at new implementation using pipes for both POSIX and windows.
It doesn't address both your considerations as this is also mostly copied from SO and it's redirecting entire stdout, but for sure it is better than previous implementation. If you still don't like it I'll prepare temporary-file based version.
Only issue I've found is #141 which is here visible as Qt 5.7+ requires C++11.


include/cucumber-cpp/internal/utils/qtCapture.hpp
```diff
@@ -0,0 +1,182 @@
...
+        setvbuf(stdout, NULL, _IONBF, 0);
+        setvbuf(stderr, NULL, _IONBF, 0);
+    }
+
+    static void BeginCapture() {
+        QMutexLocker locker(&m_mutex);
+        if (m_capturing)
+            return;
+
+        secure_pipe(m_pipe);
```
muggenhor: Where do all these `secure_XXX` variants come from? I've never encountered them before.

include/cucumber-cpp/internal/utils/qtCapture.hpp
```diff
@@ -0,0 +1,182 @@
...
+#include <QtConcurrent/QtConcurrent>
+
+#ifndef STD_OUT_FD
+#define STD_OUT_FD (fileno(stdout))
+#endif
+#ifndef STD_ERR_FD
+#define STD_ERR_FD (fileno(stderr))
+#endif
+
+class qtCapture {
```
muggenhor: None of this code is dependent on a template parameter, so it can all be moved out of this header and into a .cpp file.

include/cucumber-cpp/internal/utils/qtCapture.hpp
```diff
@@ -0,0 +1,182 @@
...
+        secure_close(m_pipe[WRITE]);
+#endif
+    }
+
+    static bool IsCapturing() {
+        QMutexLocker locker(&m_mutex);
+        return m_capturing;
+    }
+
+    static void EndCapture() {
```
muggenhor: Why is this entire class a singleton? It seems to me like the restoring of file descriptors that you're doing here is something that you would want to do in a destructor.

src/drivers/QtTestDriver.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include <QtTest/QtTest>
+
+namespace cucumber {
+namespace internal {
+
+const InvokeResult QtTestStep::invokeStepBody() {
+    QtTestObject testObject(this);
+
+    qtCapture::Init();
+    qtCapture::BeginCapture();
```
muggenhor: I'm assuming you're capturing stdout/stderr here because Qt tends to be overly verbose during execution? Is it possible to make Qt write to some alternate log sink instead?

tests/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ function(cuke_add_driver_test TEST_FILE)
     message(STATUS "Adding " ${TEST_NAME})
     add_executable(${TEST_NAME} ${TEST_FILE}.cpp)
     target_link_libraries(${TEST_NAME} cucumber-cpp-nomain ${CUKE_EXTRA_LIBRARIES} ${ARGN})
-    add_test(NAME ${TEST_NAME} COMMAND ${TEST_NAME})
+    add_test(${TEST_NAME} ${TEST_NAME})
```
muggenhor: I don't see why this change is necessary. And I do believe this is causing your test failures when running with Valgrind.

CMakeLists.txt
```diff
@@ -32,6 +34,40 @@ elseif(MSVC)
...
+            SET(CMAKE_POSITION_INDEPENDENT_CODE ON)
+        endif()
+        set(QT_LIBRARIES Qt5::Core Qt5::Widgets Qt5::Test)
+
+        #check if c++11 is nedded
+        find_program(QMAKE_EXECUTABLE NAMES qmake HINTS ${QTDIR} ENV QTDIR PATH_SUFFIXES bin)
+        execute_process(COMMAND ${QMAKE_EXECUTABLE} -query QT_VERSION OUTPUT_VARIABLE QT_VERSION)
+        if(QT_VERSION GREATER 5.6.99)
+            message(STATUS "C++11 is needed from Qt version 5.7.0, building with c++11 enabled")
+            set(CMAKE_CXX_STANDARD 11)
```
konserw: Setting this for everything fixed TCPserver tests

.travis.yml
```diff
@@ -35,9 +39,10 @@ addons:
       - google-mock
       - ninja-build
       - valgrind
+      - qtbase5-dev
 
 before_install:
-  - if [[ "${TRAVIS_OS_NAME}" = "osx" ]]; then brew update && brew install ninja; fi
+  - if [[ "${TRAVIS_OS_NAME}" = "osx" ]]; then brew update && brew install ninja && brew install qt5; fi
```
konserw: You are probably right - I have no expirience with OSX

examples/CalcQt/features/step_definitions/QtTestCalculatorQtSteps.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+static int argc = 0;
+static QApplication app(argc, 0);
+static int milliseconds = -1;
+
+int millisecondsToWait() {
+    if (milliseconds < 0)
+    {
+        char* envVariable = getenv("CALCQT_STEP_DELAY");
+        milliseconds = (0 != envVariable) ? atoi(envVariable) : 0;
```
konserw: That is not my code to be honest but I can change it

examples/CalcQt/features/step_definitions/QtTestCalculatorQtSteps.cpp
```diff
@@ -0,0 +1,72 @@
...
+    }
+    return milliseconds;
+}
+
+std::istream& operator>> (std::istream& in, QString& val) { std::string s; in >> s; val = s.c_str(); return in; }
+std::ostream& operator<< (std::ostream& out, const QString& val) { out << val.toLatin1().data(); return out; }
```
konserw: because windows

src/drivers/QtTestDriver.cpp
konserw: unfortunately only to file or stdout and we need capturing output to get results

tests/CMakeLists.txt
konserw: I'll revert it

CMakeLists.txt
muggenhor: Are they broken? Because they seem to be running fine on Travis.

AFAIK we're doing a lot of work to be C++98 compatible. Just flipping the switch like this throws away our testing of that.

examples/CalcQt/features/step_definitions/QtTestCalculatorQtSteps.cpp
muggenhor: Not even Windows has Latin-1 as default always. It's dependent on the locale, e.g. when set to Russian it has some Cyrillic code page.

Anyway, I suggest you try [toLocal8Bit()](https://doc.qt.io/qt-4.8/qstring.html#toLocal8Bit) instead. It uses whatever the locale says to use as encoding.

CMakeLists.txt
konserw: Please see issue #141 - we can discuss further there the ServerTests.

examples/CalcQt/features/step_definitions/QtTestCalculatorQtSteps.cpp
```diff
@@ -0,0 +1,72 @@
...
+GIVEN("^I just turned on the calculator$") {
+    cucumber::ScenarioScope<CalculatorWidget> calculator;
+    calculator->move(0, 0);
+    calculator->show();
+#if QT_VERSION >= 0x050000
+    QTest::qWaitForWindowExposed(calculator.get());
+#else
+    QTest::qWaitForWindowShown(calculator.get());
+#endif
+    QTest::qWait(millisecondsToWait());
```
konserw: Good idea! I've done that.

examples/CalcQt/features/step_definitions/QtTestCalculatorQtSteps.cpp
konserw: I've changed it to local8bit - let's see how it behaves on Windows

include/cucumber-cpp/internal/utils/qtCapture.hpp
konserw: Do you mean everything or just the methods definitions ?

include/cucumber-cpp/internal/utils/qtCapture.hpp
```diff
@@ -0,0 +1,182 @@
...
+#define STD_ERR_FD (fileno(stderr))
+#endif
+
+class qtCapture {
+public:
+    static void Init() {
+        // make stdout & stderr streams unbuffered
+        // so that we don't need to flush the streams
+        // before capture and after capture 
+        // (fflush can cause a deadlock if the stream is currently being
```
konserw: used?

include/cucumber-cpp/internal/utils/qtCapture.hpp
konserw: Those are private methods in that class

include/cucumber-cpp/internal/utils/qtCapture.hpp
konserw: I've converted it to normal class

konserw:
As for the c++ standard: now change to c++11 is triggered by Qt 5.8 in:
- mingw x32
- msvc x64
- OSX / GMock 1.8.0
so only 3 build out of 14 - is it acceptable ?

include/cucumber-cpp/internal/utils/qtCapture.hpp
muggenhor: The method definitions and all the `#include`s to support them.

include/cucumber-cpp/internal/utils/qtCapture.hpp
muggenhor: Ah thanks, I overlooked those.

src/drivers/QtTestDriver.cpp
```diff
@@ -0,0 +1,203 @@
...
+
+private:
+    enum PIPES { READ, WRITE };
+
+    int secure_dup(int src) {
+        int ret = -1;
+        bool fd_blocked = false;
+        do {
+             ret = dup(src);
+             fd_blocked = (errno == EINTR ||  errno == EBUSY);
```
muggenhor: This is wrong, because `errno`'s content is undefined when the previous syscall succeeded. I.e. in this case `errno` is only valid when dup's return value == -1.

src/drivers/QtTestDriver.cpp
```diff
@@ -0,0 +1,203 @@
...
+        } while (ret < 0);
+    }
+
+    void secure_close(int & fd) {
+        int ret = -1;
+        bool fd_blocked = false;
+        do {
+             ret = close(fd);
+             fd_blocked = (errno == EINTR);
+             if (fd_blocked)
```
muggenhor: `EINTR` indicates that the syscall was interrupted by a syscall and can be retried immediately. There's _no_ point in waiting first. (The same applies to other syscalls where `EINTR` is returned).

muggenhor: `EINTR` indicates that the syscall was interrupted by a syscall and can be retried immediately. There's no point in waiting first. (The same applies to other syscalls where `EINTR` is returned).

src/drivers/QtTestDriver.cpp
muggenhor: This is wrong, because `errno`'s content is undefined when the previous syscall succeeded. I.e. in this case `errno` is only valid when dup's return value == -1.

src/drivers/QtTestDriver.cpp
```diff
@@ -0,0 +1,203 @@
...
+                QThread::msleep(10);
+        } while (ret < 0);
```
muggenhor: This causes you to wait indefinitely for errors that aren't checked for explicitly. Usually you want the converse: only retry for a specific set of known error conditions.

src/drivers/QtTestDriver.cpp
```diff
@@ -0,0 +1,203 @@
...
+#define eof _eof
+#else
+#include <errno.h>
+#endif
+
+#include <cstdio>
+#include <QtConcurrent/QtConcurrent>
+
+#ifndef STD_OUT_FD
+#define STD_OUT_FD (fileno(stdout))
```
muggenhor: Please don't use this macro to hide where you're getting this file descriptor from. You're disguising it as a constant while it isn't.

src/drivers/QtTestDriver.cpp
```diff
@@ -0,0 +1,203 @@
...
+    void BeginCapture() {
+        QMutexLocker locker(&m_mutex);
+        if (m_capturing)
+            return;
+
+        secure_pipe(m_pipe);
+        m_oldStdOut = secure_dup(STD_OUT_FD);
+        m_oldStdErr = secure_dup(STD_ERR_FD);
+        secure_dup2(m_pipe[WRITE], STD_OUT_FD);
+        secure_dup2(m_pipe[WRITE], STD_ERR_FD);
```
muggenhor: This has one rather fatal flaw: it won't work if whatever you write more than the pipe's buffer can hold to either of these file descriptors. At that point it _will_ block and you're process will hang.

konserw:
> Additionally some 5 minute searching in Qt's documentation gave me the suggestion that a message handler could possibly be installed instead to capture this output. I suggest to look at that first, before trying to go this route of intercepting things: https://doc.qt.io/qt-4.8/qtglobal.html#qInstallMsgHandler

Unfortunately all QtTestLib does not rely on normal Qt debug output. they use something like this:
```
void QAbstractTestLogger::outputString(const char *msg)
{
    QTEST_ASSERT(QTest::stream);

    ::fputs(msg, QTest::stream);
    ::fflush(QTest::stream);
}
```
Where `QtTest::stream` if you do not want output to file is:
```
    if (!out) {
        QTest::stream = stdout;
        return;
}
```

konserw:
Please reevaluate temporary file based solution as qInstallHandler won't do the trick, and pipe based solution wasn't as good as I believed.

src/drivers/QtTestDriver.cpp
```diff
@@ -1,201 +1,30 @@
...
 
 const InvokeResult QtTestStep::invokeStepBody() {
-    QtTestObject testObject(this);
-
-    qtCapture capture;
-    capture.BeginCapture();
-    int returnValue = QTest::qExec(&testObject, 0, NULL);
-    capture.EndCapture();
+    QTemporaryFile file;
+    QString fileName;
```
muggenhor: Please remove this unused variable

src/drivers/QtTestDriver.cpp
```diff
@@ -1,201 +1,30 @@
...
+    if(!file.open()) {
+        return InvokeResult::failure("Unable to open temporary file needed for this test");
+    }
+    file.close();
```
muggenhor: Doesn't this or reopening it later destroy the used name? I also don't think it's necessary to close. As long as you don't read from it the buffer shouldn't be filled so you shouldn't have any problems of it containing stale data.

src/drivers/QtTestDriver.cpp
```diff
@@ -1,201 +1,30 @@
...
 
+    QtTestObject testObject(this);
+    int returnValue = QTest::qExec(&testObject, QStringList() << "test" << "-o" << file.fileName());
     if(returnValue == 0)
```
muggenhor: Add a space between `if` and `(`

src/drivers/QtTestDriver.cpp
```diff
@@ -1,201 +1,30 @@
...
         return InvokeResult::success();
     else
-        return InvokeResult::failure(capture.GetCapture());
+    {
+        file.open();
+        QTextStream ts(&file);
+        return InvokeResult::failure(ts.readAll().toLocal8Bit());
```
muggenhor: We're converting this to JSON later on. This should _definitely_ be encoded to UTF-8, _not_ to the local character set. Let cucumber take care of doing that conversion when it wishes to display its output.

src/drivers/QtTestDriver.cpp
konserw: It doesn't destroy used name, but it turns out to be unnecessary to close and reopen.

src/drivers/QtTestDriver.cpp
konserw: done

src/drivers/QtTestDriver.cpp
konserw: done

konserw:
Closing this as I've created another revision of this PR #165.

Issue #123, opened 2016-09-16
Code Coverage?


dbzVT8:
Ruby-based Cucumber uses the Rake task to perform code coverage. Any interest in implementing code coverage for cucumber-cpp?

paoloambrosio:
It could be a good idea. What tools are you thinking about?

dbzVT8:
No I do not have any recommendations for a specific tool because I do not know exactly how best to integrate it into cucumber-cpp. I am rather new to Cucumber in general and I have next to no Ruby experience so looking at how code coverage is implemented into Ruby-based Cucumber doesn't help me gain insight into how it can be done on the cpp side.  However, without knowing the best way to integrate, I would say the two tools that do come to mind are maybe COVTOOL or CppUnit.

PR #143, opened 2017-04-05
Coveralls based coverage testing

konserw:
Code coverage solution based on gcov and Coveralls service: https://coveralls.io/github/konserw/cucumber-cpp
Of course badge and api key is to be replaced.
Fixes issue #123

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,10 +14,7 @@ set(GMOCK_SRC_DIR "" CACHE STRING "Google Mock framework sources path (otherwise
 set(GMOCK_VER "1.7.0" CACHE STRING "Google Mock framework version to be used")
 option(VALGRIND_TESTS "Run tests within Valgrind" OFF)
 
-set(CMAKE_MODULE_PATH
-    ${CMAKE_MODULE_PATH} 
-    ${PROJECT_SOURCE_DIR}/cmake/modules
-)
+list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/module)
 
 #
 # Generic Compiler Flags
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,7 +14,7 @@ set(GMOCK_SRC_DIR "" CACHE STRING "Google Mock framework sources path (otherwise
 set(GMOCK_VER "1.7.0" CACHE STRING "Google Mock framework version to be used")
 option(VALGRIND_TESTS "Run tests within Valgrind" OFF)
 
-list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/module)
+list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
 
 #
 # Generic Compiler Flags
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,8 +13,18 @@ set(CUKE_ENABLE_EXAMPLES OFF CACHE BOOL "Enable the examples")
 set(GMOCK_SRC_DIR "" CACHE STRING "Google Mock framework sources path (otherwise downloaded)")
 set(GMOCK_VER "1.7.0" CACHE STRING "Google Mock framework version to be used")
 option(VALGRIND_TESTS "Run tests within Valgrind" OFF)
+option(COVERALLS "Run coverage tests for Coveralls" OFF)
 
-set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
+set(CMAKE_MODULE_PATH
+    ${CMAKE_MODULE_PATH} 
+    ${PROJECT_SOURCE_DIR}/cmake/modules
+    ${PROJECT_SOURCE_DIR}/coveralls-cmake/cmake
+)
+
+if(COVERALLS)
+	include(Coveralls)
+	coveralls_turn_on_coverage()
+endif()
 
 #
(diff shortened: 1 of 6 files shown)


muggenhor:
This is a nice idea, but instead of hacking this into our regular build system I would prefer an approach that changes only the CI system. With only very minimal changes to the build system (at the most I would expect the build system to produce a list of relevant source files).

What I would expect instead:

1. modifying the CI script to add the desired [instrumentation options](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html) to CMake's invocation (probably: `-DCMAKE_CXX_FLAGS='--coverage' -DCMAKE_LINK_INTERFACE_LIBRARIES='gcov'`).
2. adding a post-processing step that collects the desired output after test execution and transmits it to coveralls

The first bit is probably just adding this to travis.sh's CMake invocation:
```sh
${COVERALLS_SERVICE_NAME:+${COVERALLS_REPO_TOKEN:+-DCMAKE_CXX_FLAGS='--coverage' -DCMAKE_LINK_INTERFACE_LIBRARIES='gcov'}}
```

The second bit is more involved but should be similar to this pseudo code:
```sh
if [ -n "${COVERALLS_SERVICE_NAME:-}" -a "${COVERALLS_REPO_TOKEN:-}" ]; then
    # Let gcov extract coverage data
    cmake -E make_directory build/gcov
    (
        cd build/gcov
        for f in `find .. -name '*.gcda' -type f`; do
            gcov -p -o "$(dirname "$f")" "$f"
        done
    ) > /dev/null

    # Translate _relevant_ gcov files to a format coveralls can consume
    for f in build/gcov/*\#cucumber-cpp\#src\#* build/gcov/*\#include\#cucumber-cpp\#*; do
        convert_from_gcov_to_coveralls "$f"
    done

    transmit_coverage_data_to_coveralls
fi
```

I expect none of that needs changes to the build system. The only exception would be if we would want to have some of the build system's info available to us to determine which files are relevant to display coverage for (instead of the very simple, but effective, glob patterns I listed).

konserw:
@muggenhor @paoloambrosio @aslakhellesoy - could you create cucumber account on coveralls ?

src/CMakeLists.txt
```diff
@@ -33,6 +33,9 @@ if(CMAKE_EXTRA_GENERATOR OR MSVC_IDE)
     list(APPEND CUKE_SOURCES ${CUKE_HEADERS})
 endif()
 
+string(REGEX REPLACE "([^;]+)" "${CMAKE_CURRENT_SOURCE_DIR}/\\1" COVER_SOURCES "${CUKE_SOURCES}")
+file(WRITE sourceFileList "${COVER_SOURCES}" )
```
konserw: It was just a thing I was playing with and even deleted from my repo - just forgot pushing deletion commit here

CMakeLists.txt
```diff
@@ -14,7 +14,10 @@ set(GMOCK_SRC_DIR "" CACHE STRING "Google Mock framework sources path (otherwise
 set(GMOCK_VER "1.7.0" CACHE STRING "Google Mock framework version to be used")
 option(VALGRIND_TESTS "Run tests within Valgrind" OFF)
 
-set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
+set(CMAKE_MODULE_PATH
+    ${CMAKE_MODULE_PATH} 
+    ${PROJECT_SOURCE_DIR}/cmake/modules
+)
```
konserw: Ok

travis.sh
```diff
@@ -7,27 +7,35 @@ bundle install
 CTEST_OUTPUT_ON_FAILURE=ON
 export CTEST_OUTPUT_ON_FAILURE
 
+export COVERALLS_SERVICE_NAME=travis-ci
+export COVERALLS_REPO_TOKEN=ThdhpdagGJJVJj0KFGdSgdpzRLALbNAIg
```
konserw: We can try without token

CMakeLists.txt
```diff
@@ -14,7 +14,7 @@ set(GMOCK_SRC_DIR "" CACHE STRING "Google Mock framework sources path (otherwise
 set(GMOCK_VER "1.7.0" CACHE STRING "Google Mock framework version to be used")
 option(VALGRIND_TESTS "Run tests within Valgrind" OFF)
 
-set(CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
+list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake/modules)
```
muggenhor: I think this is causing your build failures. I guess `list(APPEND)` doesn't work for variables not yet defined.

muggenhor:
So it looks like POSTing to coveralls works, but I don't think it manages to extract the coverage data:
>  **0** of **0** relevant lines covered (**NaN%**)

konserw:
I don't understand this - build from the same branch/commit is fine on my account:
https://coveralls.io/github/konserw/cucumber-cpp

muggenhor:
Well at least it's not the absence of the API token that's at fault, when that's included this is the result: https://coveralls.io/builds/12854344

konserw:
> I don't understand this - build from the same branch/commit is fine on my account:
https://coveralls.io/github/konserw/cucumber-cpp

Please disregard that comment - I was looking at wrong branch. With this branch I've also got 0 relevant lines.

konserw:
I don't know how to fix this implementation. I'll create new pull request with different approach.

muggenhor:
FYI: this is continued in #162

PR #146, opened 2017-04-08
Better contributing guidelines and GitHub templates for issues and PRs

paoloambrosio:
- Written down in `CONTRIBUTING.md` some instructions for maintainers
- Added GitHub templates for PRs and issues

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,50 @@
+<!-- These sections are meant as guidance for you, to help you give the kind of information we'll need to help with your issue. If a section  doesn't seem to fit, just skip it.
+
+This is not a support or discussion forum. If you have a question, please ask it on [The Cukes Google Group](http://groups.google.com/group/cukes) or on [Gitter](https://gitter.im/cucumber/).
+
+In general: Please provide as much information as you can to help us solving your problem -->
+
+## Summary
+
+<!--- Provide a general summary description of the issue -->
+
+## Expected Behavior
+
+<!--- If you're describing a bug, tell us what should happen -->
+<!--- If you're suggesting a change/improvement, tell us how it should work -->
+<!--- Feel free to use Given / Then / Then if that helps, but please add some plain-language context too -->
+
+## Current Behavior
+
+<!--- If describing a bug, tell us what happens instead of the expected behavior -->
+<!--- If suggesting a change/improvement, explain the difference from current behavior -->
+
+<!--- If you have got some output place it in the code block below. Otherwise remove it. -->
+~~~
+~~~
+
+## Possible Solution
+
+<!--- Not obligatory, but suggest a fix/reason for the bug, -->
+<!--- or ideas how to implement the addition or change -->
+
+## Steps to Reproduce (for bugs)
+
+<!--- Provide a link to a live example, or an unambiguous set of steps to -->
+<!--- reproduce this bug. Include code to reproduce, if relevant -->
+1.
+2.
+3.
+4.
+
+## Context & Motivation
+
+<!--- How has this issue affected you? What are you trying to accomplish? -->
+<!--- Providing context helps us come up with a solution that is most useful in the real world -->
+
+## Your Environment
+
+<!--- If you're reporting a bug, include as many relevant details about the environment you experienced the bug in -->
+* Version used:
+* Operating System and version:
+* Link to your project:
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -12,7 +12,7 @@ In general: Please provide as much information as you can to help us solving you
 
 <!--- If you're describing a bug, tell us what should happen -->
 <!--- If you're suggesting a change/improvement, tell us how it should work -->
-<!--- Feel free to use Given / Then / Then if that helps, but please add some plain-language context too -->
+<!--- Feel free to use Given / When / Then if that helps, but please add some plain-language context too -->
 
 ## Current Behavior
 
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Summary
+
+<!--- Provide a general summary description of the issue -->
+
+## Expected Behavior
+
+<!--- If you're describing a bug, tell us what should happen -->
+<!--- If you're suggesting a change/improvement, tell us how it should work -->
+<!--- Feel free to use Given / Then / Then if that helps, but please add some plain-language context too -->
```
muggenhor: I think you meant `Given / When / Then` ("When" instead of "Then").

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
...
+<!--- Please describe in detail how you tested your changes. -->
+<!--- Include details of your testing environment, tests ran to see how -->
+<!--- your change affects other areas of the code, etc. -->
+
+## Types of changes
+
+<!--- What types of changes does your code introduce? Put an `x` in all the boxes that apply: -->
+- [ ] Bug fix (non-breaking change which fixes an issue).
+- [ ] New feature (non-breaking change which adds functionality).
+- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
```
muggenhor: I suggest changing "as expected" to "the same". Otherwise you're asking the submitter to guess about random other people's expectations.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Checklist:
+
+<!--- Go over all the following points, and put an `x` in all the boxes that apply. -->
+<!--- If you're unsure about any of these, don't hesitate to ask. We're here to help! -->
+- [ ] It is my own work, its copyright is implicitly assigned to the project and no substantial part of it has been copied from other sources (including [Stack Overflow](https://stackoverflow.com/)). In rare occasions this is acceptable, like in CMake modules where the original copyright information should be kept.
+- [ ] I'm using the same code standards as the existing code (indentation, spacing, variable naming, ...).
```
muggenhor: At my work we've actually added a CI step that uses clang-format to detect whether differences, deviating from the described coding style, have been introduced and depending on context automatically reformats and resubmits, or rejects the change. I think something like that might be of more value than asking people to tick this box.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
...
+- [ ] I've added tests for my code.
```
muggenhor: As a checklist item I think this has value (both for the submitter and reviewers). I'm less sure about the value for publicly ticking that box though.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
...
+- [ ] My change requires a change to the documentation and I have updated the documentation accordingly.
```
muggenhor: These are two separate statements and should be separate check list items too IMO. E.g.:
- [ ] I have verified that my change requires no change to the documentation.

This splits the verification task (producing no committable artifacts) from the documentation task.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,39 @@
...
+- [ ] My branch has been rebased to master, keeping only relevant commits.
```
muggenhor: I think it's okay too if it can be merged to master without conflicts (other than perhaps `HISTORY.md`). I don't believe it's necessary to rebase to the tip of master constantly if it doesn't address any merge conflicts. Although that's more a 'standard of behaviour' nuance that I'm not sure we need to (or want to) express in this check list.

CONTRIBUTING.md
```diff
@@ -37,14 +37,38 @@ Here is an [Example](https://github.com/cucumber/bool/pull/12) of this process i
...
+    discussion has occurred for any breaking change
+  - The branch does not need rebasing or squashing of commits
+- To merge:
+  - Follow the command line instructions on GitHub
+  - If it is either a new feature or a bugfix, specify the `--no-commit`
+    flag when merging and add a line to `HISTORY.md` following the
+    current convention. Add this file to the changes to be committed and
+    commit the merge.
+  - Commit message should follow the current convention:
+    `Merge #42 'Description of the change usually from the PR description'`
```
muggenhor: Is this the message github auto generates when you use the web interface for merging? If so then I think it is helpful to mention that fact here.

CONTRIBUTING.md
```diff
@@ -37,14 +37,38 @@ Here is an [Example](https://github.com/cucumber/bool/pull/12) of this process i
...
+
+### Do a release
+
+- Release commit (e.g. [fdf8a5c](https://github.com/cucumber/cucumber-cpp/commit/fdf8a5c4ef4c51dfa7ea82077f706414a4c6322d)):
+  - Change `HISTORY.md` renaming the "In Git" section with the
+    release number and date
+  - Commit with message `Updated history file for the X.Y release`
+  - Tag this commit with `vX.Y`
```
muggenhor: Should this be an annotated tag or just a tag reference (unannotated) one?

paoloambrosio: You are right pointing it out. I made a mess myself (with a mix of them) but they should be annotated.

.github/PULL_REQUEST_TEMPLATE.md
muggenhor: I've created task #151 for this. It shouldn't be too much work, I'll look at it when I have some time.

muggenhor:
@paoloambrosio I think this looks good right now to merge in. Do you agree?

PR #147, opened 2017-05-25
Support building as a shared library on Windows

nre-ableton:
On Windows, it is necessary to add some __declspec() calls to expose
some symbols with different linkage types. This commit adds the
CUKE_API_ macro, which makes it possible for users to easily build
cucumber-cpp on Windows either as a static or dynamic library by
defining either CUKE_LINKED_AS_SHARED_LIBRARY or
CUKE_CREATE_SHARED_LIBRARY at build time.


include/cucumber-cpp/internal/CukeDll.hpp
```diff
@@ -0,0 +1,14 @@
+#ifndef CUKE_CUKEDLL_HPP_
```
muggenhor: I'd appreciate it if you could use CMake's [GenerateExportHeader](https://cmake.org/cmake/help/v3.0/module/GenerateExportHeader.html) module, or at least make this header match the output that would be generated by that module much closer. To produce this output to get an example, add this to `src/CMakeLists.txt` and look for the generated files in `$BUILDDIR/src/`:
```cmake
include(GenerateExportHeader)

generate_export_header(cucumber-cpp-nomain)
generate_export_header(cucumber-cpp)
```

That should pave the way for symbol visibility on non-Windows OSes too, without having to redo the Windows part later on.

Additionally: to reflect that, please rename this header to not mention DLLs, but the fact that you're declaring which symbols are to be exported. E.g. `CukeExport.h`.

include/cucumber-cpp/internal/CukeDll.hpp
```diff
@@ -0,0 +1,14 @@
...
+#define CUKE_CUKEDLL_HPP_
+
+#ifndef CUKE_API_
```
muggenhor: Don't do this check, it should be considered a bug (and thus reported by the compiler) if it's already defined.

include/cucumber-cpp/internal/CukeDll.hpp
```diff
@@ -0,0 +1,14 @@
...
+#if CUKE_LINKED_AS_SHARED_LIBRARY
```
muggenhor: Just like GenerateExportHeader, please:

* wrap both the export and import case in `#ifndef CUCUMBER_CPP_STATIC_DEFINE` to detect when we're building or using a static library.
* distinguish the export from the import case by the presence of the macros `cucumber_cpp_EXPORTS` and `cucumber_cpp_nomain_EXPORTS` (CMake defines those only while building objects for the mentioned target)

include/cucumber-cpp/internal/defs.hpp
```diff
@@ -1,5 +1,6 @@
 #include "step/StepManager.hpp"
 #include "hook/HookRegistrar.hpp"
 #include "ContextManager.hpp"
+#include "CukeDll.hpp"
```
muggenhor: Please follow standard "[include-what-you-use](https://include-what-you-use.org/)" rules and _don't_ include what you don't use (in the same file).

include/cucumber-cpp/internal/CukeDll.hpp
```diff
@@ -0,0 +1,14 @@
...
+#define CUKE_API_ __declspec(dllimport)
+#elif CUKE_CREATE_SHARED_LIBRARY
+#define CUKE_API_ __declspec(dllexport)
+#else
+#define CUKE_API_
```
muggenhor: FYI: GenerateExportHeader would call this `CUCUMBER_CPP_EXPORT` (and for Windows `#define` that to be dllimport when importing).

include/cucumber-cpp/internal/CukeEngineImpl.hpp
```diff
@@ -13,7 +13,7 @@ namespace internal {
  * Currently it is a wrapper around CukeCommands. It will have its own
  * implementation when feature #31 is complete.
  */
-class CukeEngineImpl : public CukeEngine {
+class CUKE_API_ CukeEngineImpl : public CukeEngine {
```
muggenhor: You forgot to include the header. It probably builds because of an indirect include, but please don't depend on that ([include-what-you-use](https://include-what-you-use.org/): directly in the file where it's used).

include/cucumber-cpp/internal/CukeDll.hpp
nre-ableton: Sure, I can fix `CMakeLists.txt` to use `GenerateExportHeader`. It seems to me though that it would make `CukeDll.hpp` (or `CukeExport.h`, or however it should be named) to be obsolete. If it's ok with you, I'll just remove this file and add the necessary `include_directories` to CMake so that it finds the generated `cucumber-cpp_export.h` file.

include/cucumber-cpp/internal/CukeDll.hpp
nre-ableton: Ok, good point. However I will instead remove this file (see above comment).

include/cucumber-cpp/internal/defs.hpp
nre-ableton: Good point, removed.

include/cucumber-cpp/internal/CukeEngineImpl.hpp
nre-ableton: Fixed

nre-ableton:
@muggenhor thanks for the feedback! I force pushed this branch with some fixes and also cleaned up the commit message to make it more platform-generic. `CukeDll.hpp` has been completely removed, and now the project relies on the auto-generated header from CMake's `GenerateExportHeader` module instead.

include/cucumber-cpp/internal/CukeDll.hpp
muggenhor: Replacing completely with the generated done is okay with me. Please make that `target_include_directories` so that transitive build requirements work properly.

nre-ableton: @muggenhor ah, I forgot about that, added a fixup just now for that.

src/CMakeLists.txt
```diff
@@ -33,8 +35,23 @@ if(CMAKE_EXTRA_GENERATOR OR MSVC_IDE)
...
+
+add_library(cucumber-cpp-nomain ${CUKE_LINKAGE} ${CUKE_SOURCES})
+add_library(cucumber-cpp ${CUKE_LINKAGE} ${CUKE_SOURCES} main.cpp)
+
+set_target_properties(cucumber-cpp cucumber-cpp-nomain PROPERTIES
+    DEFINE_SYMBOL cucumber_cpp_EXPORTS
+)
+generate_export_header(cucumber-cpp)
+
+include_directories(${CMAKE_CURRENT_BINARY_DIR})
```
muggenhor: This should no longer be necessary, `target_include_directories` replaces it.

src/CMakeLists.txt
```diff
@@ -33,8 +35,22 @@ if(CMAKE_EXTRA_GENERATOR OR MSVC_IDE)
...
+    set(CUKE_LINKAGE STATIC)
+endif()
+
+add_library(cucumber-cpp-nomain ${CUKE_LINKAGE} ${CUKE_SOURCES})
+add_library(cucumber-cpp ${CUKE_LINKAGE} ${CUKE_SOURCES} main.cpp)
+
+set_target_properties(cucumber-cpp cucumber-cpp-nomain PROPERTIES
+    DEFINE_SYMBOL cucumber_cpp_EXPORTS
+)
+generate_export_header(cucumber-cpp)
```
paoloambrosio: What about using the `EXPORT_FILE_NAME` parameter of `GENERATE_EXPORT_HEADER` (if I read correctly) to match the current file naming conventions for include files? `#include "cucumber-cpp_export.h"` does not read well

src/CMakeLists.txt
```diff
@@ -33,8 +35,25 @@ if(CMAKE_EXTRA_GENERATOR OR MSVC_IDE)
...
+
+add_library(cucumber-cpp-nomain ${CUKE_LINKAGE} ${CUKE_SOURCES})
+add_library(cucumber-cpp ${CUKE_LINKAGE} ${CUKE_SOURCES} main.cpp)
+
+set_target_properties(cucumber-cpp cucumber-cpp-nomain PROPERTIES
+    DEFINE_SYMBOL cucumber_cpp_EXPORTS
+)
+
+generate_export_header(cucumber-cpp
+    EXPORT_FILE_NAME "CukeExport.hpp"
```
muggenhor: In order to place this in the same, relative, include directory structure, I think we should use `EXPORT_FILE_NAME cucumber-cpp/internal/CukeExport.h` (this will affect the required include directives which will need to change to `#include <cucumber-cpp/internal/CukeExport.h>`). This would make library installs place the headers all in a single location.

include/cucumber-cpp/internal/ContextManager.hpp
```diff
@@ -16,7 +18,7 @@ namespace internal {
 
 typedef std::vector<shared_ptr<void> > contexts_type;
 
-class ContextManager {
+class CUCUMBER_CPP_EXPORT ContextManager {
```
muggenhor: I think it's possible to export only the member function `purgeContexts` and add a non-template overload for `addContext` and export that too.

But this can also be done after this PR.

include/cucumber-cpp/internal/CukeEngine.hpp
```diff
@@ -59,7 +61,7 @@ class PendingStepException : public InvokeException {
  * It uses standard types (as much as possible) to be easier to call.
  * Returns standard types if possible.
  */
-class CukeEngine {
+class CUCUMBER_CPP_EXPORT CukeEngine {
```
muggenhor: I believe all of the members of this class are used internally only, with the exception of the constructor and destructor. So if you explicitly declare those and add (empty) implementations to the .cpp file then exporting just those should be enough.

I know on GNU/Linux that would work. Could you verify that that works on MSVC too? I.e. only exporting some members of a class without exporting the entire class?

muggenhor:
Oh, before I forget. Could you please add (or modify) an AppVeyor build that _will_ fail to build at least one of the tests or examples if one of the symbols you currently export doesn't get exported? I'd appreciate it if you could demonstrate that failure with an intermediate commit.

This to ensure we don't accidentally break this in the future.

nre-ableton:
@muggenhor Sorry for the delay on finishing up this PR, I had some other work that I had to prioritize in the last few weeks. Anyways, sure, I can adapt the AppVeyor build configuration. I'll be creating another PR so that I can make sure that it fails/succeeds as expected. Please ignore that PR; I'll ping you back in this one once everything is ready to be reviewed again.

include/cucumber-cpp/internal/CukeEngine.hpp
nre-ableton: AFAIK MSVC should support exporting individual methods and not the entire class. I can test this with `dumpbin /EXPORTS` (tho I'm not on my Windows machine now, so I'll comment again once I've verified that).

nre-ableton:
@muggenhor Thanks for dusting off and reviving this PR. Do you need any help from me to get it ready to merge? Namely, should I rebase it and fix the merge conflicts?

muggenhor:
I have a rebase ready as e9d074a47 already. If CI succeeds for d13440a2 (a commit for CI only that I don't intend to merge) then I intend to push these to this PR. I would appreciate it if you could confirm that it still works as intended for you at that point. That'll probably be about 2 hours from now.

nre-ableton:
@muggenhor I don't have time today to test these changes, but I'm sure they'll be fine. 😄 I did a quick look at the code and I'm sure that it'll work fine for us. We don't use CMake internally and most of the source changes are pretty trivial. Go ahead and merge!

muggenhor:
I intend to merge this tomorrow. It now works on all CI, just minor stuff turned out to be the more difficult things.

muggenhor:
This PR has been merged into master with 183c79d. Thanks for your change!

PR #148, opened 2017-05-25
Use HTTPS protocol instead of git for submodules

nre-ableton:
HTTPS is the protocol least likely to be blocked by a corporate
firewall.


muggenhor:
Thanks for this. Not sure why this hasn't yet bitten me personally at work yet.

PR #149, opened 2017-05-25
Add a step-matching hook

nre-ableton:
The motivation behind this commit is that we get too many ambiguous
step matchings because we use (.+) as view expressions. What's a view
and what isn't is only determined at runtime (through the view
evaluators which register at runtime), so we can't describe it in a
regular expression.

With this feature, we have a say in whether or not a step should
match, by providing our own matching hook. We use regex's named
capture mechanism to activate our own matching algorithm.

Credit to @stk-ableton for the original patch.

### DIFF
--- a/include/cucumber-cpp/internal/hook/HookRegistrar.hpp
+++ b/include/cucumber-cpp/internal/hook/HookRegistrar.hpp
@@ -63,6 +63,7 @@ class HookRegistrar {
 public:
     typedef std::list< boost::shared_ptr<Hook> > hook_list_type;
     typedef std::list< boost::shared_ptr<AroundStepHook> > aroundhook_list_type;
+    typedef bool(*StepMatchingHook)(const boost::cmatch&);
 
     virtual ~HookRegistrar();
 
@@ -84,6 +85,9 @@ public:
     void addAfterAllHook(const boost::shared_ptr<AfterAllHook>& afterAllHook);
     void execAfterAllHooks();
 
+    void setStepMatchingHook(StepMatchingHook hook);
+    bool execStepMatchingHook(const boost::cmatch& originalMatch);
+
 private:
     void execHooks(HookRegistrar::hook_list_type &hookList, Scenario *scenario);
 
@@ -94,6 +98,7 @@ protected:
     hook_list_type& afterStepHooks();
     hook_list_type& afterHooks();
     hook_list_type& afterAllHooks();
+    StepMatchingHook& stepMatchingHook();
 };
 
 
--- a/src/HookRegistrar.cpp
+++ b/src/HookRegistrar.cpp
@@ -95,7 +95,6 @@ void HookRegistrar::execAfterHooks(Scenario *scenario) {
     execHooks(afterHooks(), scenario);
 }
 
-
 void HookRegistrar::execHooks(HookRegistrar::hook_list_type &hookList, Scenario *scenario) {
     for (HookRegistrar::hook_list_type::iterator hook = hookList.begin(); hook != hookList.end(); ++hook) {
         (*hook)->invokeHook(scenario, NULL);
@@ -126,6 +125,19 @@ void HookRegistrar::addAfterAllHook(const boost::shared_ptr<AfterAllHook>& after
 
 void HookRegistrar::execAfterAllHooks() {
     execHooks(afterAllHooks(), NULL);
+}
+
+void HookRegistrar::setStepMatchingHook(StepMatchingHook hook) {
+    stepMatchingHook() = hook;
+}
+
+bool HookRegistrar::execStepMatchingHook(const boost::cmatch& originalMatch) {
+    return stepMatchingHook() == nullptr || stepMatchingHook()(originalMatch);
+}
+
+HookRegistrar::StepMatchingHook& HookRegistrar::stepMatchingHook() {
+    static StepMatchingHook stepMatchingHook = nullptr;
+    return stepMatchingHook;
 }
 
 
--- a/src/HookRegistrar.cpp
+++ b/src/HookRegistrar.cpp
@@ -132,11 +132,11 @@ void HookRegistrar::setStepMatchingHook(StepMatchingHook hook) {
 }
 
 bool HookRegistrar::execStepMatchingHook(const boost::cmatch& originalMatch) {
-    return stepMatchingHook() == nullptr || stepMatchingHook()(originalMatch);
+    return stepMatchingHook() == NULL || stepMatchingHook()(originalMatch);
 }
 
 HookRegistrar::StepMatchingHook& HookRegistrar::stepMatchingHook() {
-    static StepMatchingHook stepMatchingHook = nullptr;
+    static StepMatchingHook stepMatchingHook = NULL;
     return stepMatchingHook;
 }
 
--- a/src/Regex.cpp
+++ b/src/Regex.cpp
@@ -1,4 +1,5 @@
 #include <cucumber-cpp/internal/utils/Regex.hpp>
+#include <cucumber-cpp/internal/hook/HookRegistrar.hpp>
 #include <boost/make_shared.hpp>
 
 namespace cucumber {
@@ -26,7 +27,8 @@ boost::shared_ptr<RegexMatch> Regex::find(const std::string &expression) const {
 
 FindRegexMatch::FindRegexMatch(const boost::regex &regexImpl, const std::string &expression) {
     boost::cmatch matchResults;
-    regexMatched = boost::regex_search(expression.c_str(), matchResults, regexImpl, boost::regex_constants::match_extra);
+    regexMatched = boost::regex_search(expression.c_str(), matchResults, regexImpl, boost::regex_constants::match_extra)
+                   && HookRegistrar().execStepMatchingHook(matchResults);
     if (regexMatched) {
         for (boost::cmatch::size_type i = 1; i < matchResults.size(); ++i) {
             RegexSubmatch s;


paoloambrosio:
I don't understand the problem either, and agree with @muggenhor that internal includes should not be used (the name was meant to communicate that!).

paoloambrosio:
@nre-ableton still waiting for an explanation of why you need this, in the form of what is impossible to do with standard Cucumber. A technical solution without a business case is never good, even for the software Cucumber is meant to test ;-)

It could also be a good idea to write in the [Cucumber Discussion Group](http://groups.google.com/group/cukes) to get responses from the wider community.

nre-ableton:
@paoloambrosio yeah, sorry for pushing a fixup without answering any of the larger questions at hand. 🙂 
I'm submitting this patch on behalf of other colleagues in my company, so I had to do some additional investigating for the motivation, as I'm also rather new to the world of cucumber.

The idea here is that we want to add support for view expressions, so that we can filter steps based on runtime information, and thus enable/disable these steps at runtime. A similar feature can be found in cucumber-ruby: https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/extending_cucumber/custom_filter.feature

I'll see if I can find/adapt an example test snippit from our cucumber tests to shed some more light on this.

Regarding the public/private API, I can completely understand your point here, but I am not entirely certain how to refactor this patch to make it part of the public API. When you say that the headers should be split to expose less functionality, do you think that this should be done as a prerequisite to this patch?

muggenhor:
@nre-ableton
> Regarding the public/private API, I can completely understand your point here, but I am not entirely certain how to refactor this patch to make it part of the public API. When you say that the headers should be split to expose less functionality, do you think that this should be done as a prerequisite to this patch?

No, splitting the headers would only prevent confusion such as this in the future, it's not a prerequisite.

paoloambrosio:
Don't think that Cucumber-Ruby feature does what this PR implements. It filters scenarios before they are run and not step definitions being matched. Given that the scenario execution in Cucumber-CPP is run by Cucumber-Ruby, that functionality should be available on the Ruby side.

I have also never heard of "view expressions", and even googling didn't help.

I suspect there might be a better solution to your problem, so my advice would be for one of your colleagues to post a message in the Google Group to seek advice (from people with a lot of experience dealing with any kind of software).

toh-ableton:
Hi, let me try to explain a little what we're trying to achieve. In our test suite, we have steps like
```gherkin
Then the Back To Arrangement button should be lit
Then the Re-Enable Automation button should be lit
Then the Session Record button should be lit
# and so on...
```
The regex for the step definition looks somewhat like this:
```c++
Then("^(.+) should be (lit|unlit)$")
{
    REGEX_PARAM(string, ViewExp);
    REGEX_PARAM(string, ShouldBeLitString);
    // ...
}
```
Note that we capture the parameter `ViewExp` using `(.+)`, which is super generic. The reason for this is that we don't want to have to list every possible button in the regexp (because there are so many of them). Unfortunately, this sometimes matches too much because we have other steps that look like this:
```gherkin
Then shortly afterwards, the Back To Arrangement button should be lit
```
which should be matched by this step definition:
```ruby
Then /^shortly afterwards, (.+[^:])$/ do |action|
    step "I wait some time"
    step action
end
```
Because we're matching `ViewExp` using `(.+)` above, this step matches both step definitions, even though it shouldn't match the first one because `shortly afterwards, the Back To Arrangement button` isn't a button (`the Back To Arrangement button` is a button).

Our solution internally to this problem is this patch, which allows to filter out step definitions based on runtime information in cases where regular expressions aren't expressive enough.

`cucumber-ruby` provides a slightly different extension point for problems like ours. `cucumber-ruby` implements its main loop as a pipeline of [`Filter`s](https://github.com/cucumber/cucumber-ruby-core/blob/master/lib/cucumber/core/filter.rb) that process test cases. A filter can filter out test cases (like the [`TagFilter`](https://github.com/cucumber/cucumber-ruby-core/blob/master/lib/cucumber/core/test/filters/tag_filter.rb)) or retry a failed test case with the [`Retry` filter](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/filters/retry.rb) or provide step definitions like the [`ActivateSteps` filter](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/filters/activate_steps.rb).

The main filter pipeline in `cucumber-ruby`'s [Runtime](https://github.com/cucumber/cucumber-ruby/blob/master/lib/cucumber/runtime.rb#L255) provides an extension point for custom filters. The [example from the `Feature`](https://relishapp.com/cucumber/cucumber/docs/extending-cucumber/custom-filter) that tests this feature does exactly that: even though there's no step definition for `Given my special step`, the `MakeAnythingPass` filter provides an [empty step definition](https://github.com/cucumber/cucumber-ruby/blob/master/features/docs/extending_cucumber/custom_filter.feature#L17).

Now the wire protocol doesn't know test cases, it only knows about steps, hooks (and snippets). So the closest thing in C++ to a custom filter on test cases in my opinion would be to allow custom filters on step definitions (and possibly hooks). This PR is maybe a first stab at this, but I'm not sure if you as `cucumber-cpp` maintainers are interested in such a feature and what it should look like.

If you'd like to, we could take this PR to discuss the desired design for such an extension mechanism. On the other hand, if you feel this adds too much complexity or you don't like this feature for some other reason, we'd also be fine with maintaining this patch in our private fork.

muggenhor:
It seems to that in your examples you very well could distinguish the button names: they're all prefixed with "the". Alternatively you could do what we do at our company: wrap names from enumerations like this in quotes (or other markers) like this:
```cucumber
Given the "Session Initiation" button
```

```cpp
GIVEN ("the \"(.*?)\" button") {
  REGEX_PARAM (string, button_name);
}
```

paoloambrosio:
I understand how you use filters in Cucumber-Ruby now. From what you wrote, it doesn't look like they are needed though.

I'm sure you understand that one of the responsibilities of a maintainer is to prevent feature bloating. Every feature that is not present in most Cucumber implementation and is not C++ specific is a potential candidate for that. Over the years, I've seen so many times users abusing every bit of flexibility in Cucumber and putting themselves in a corner. I'd rather implement a missing functionality (this PR or something else we agree) or find that it's not necessary, other than you maintaining a separate fork :-)

Let's see if any of the following ideas are a good fit.

**Eventually / test probes**

Is "shortly afterwards" a specification or a technical requirement of the test automation? Given how it is phrased, it looks technical to me. In that case it shouldn't be in the Feature file.

Let's assume that some checks take longer, but they are idempotent. A common pattern in integration testing is to poll and fail only after some timeout (good reference for this is [Growing Object-Oriented Software, Guided by Tests](https://www.amazon.com/Growing-Object-Oriented-Software-Guided-Tests/dp/0321503627/)). If all "view checks" are implemented like that, there should be no need for a specific "shortly afterwards" version. This has the big advantage of slowing down checks as much as it is needed and no more, compared to waiting before checking because you know it's going to take some time.

**Surround with quotes**

You could also surround specific names (e.g. button names) with quotes:
```
Then the Back To Arrangement button should be lit
Then shortly afterwards, the Back To Arrangement button should be lit
```
becomes
```
Then the "Back To Arrangement" button should be lit
Then shortly afterwards, the "Back To Arrangement" button should be lit
```

**Fix the regexes**

If all of this is not appealing to you, here are two regexes that should fit your case:
 - `^shortly afterwards, ([^,]+) should be (lit|unlit)$/`
 - `^([^,]+) should be (lit|unlit)$`

I've used the Ruby interpreter to try them:
```
sr = /^shortly afterwards, the (.+) should be (lit|unlit)$/
sr.match("shortly afterwards, the Back To Arrangement button should be lit")
 => #<MatchData "shortly afterwards, the Back To Arrangement button should be lit" 1:"Back To Arrangement button" 2:"lit"> 
sr.match("the Back To Arrangement button should be lit")
 => nil

r = /^the (.+) should be (lit|unlit)$/
r.match("shortly afterwards, the Back To Arrangement button should be lit")
 => nil
r.match("the Back To Arrangement button should be lit")
 => #<MatchData "the Back To Arrangement button should be lit" 1:"Back To Arrangement button" 2:"lit">
```
Hope you don't mind me removing the "the", but of course there are ways to keep it if it's really important:
 - `^shortly afterwards, ([^,]+) should be (lit|unlit)$`
 - `^([^,]+) should be (lit|unlit)$`

toh-ableton:
Hey @paoloambrosio, thanks for your detailed response, I appreciate it.

Thanks in particular for suggesting other workarounds. Just to give you a little more context: we have 85 different expressions like `the Back To Arrangement button`, which are used in 29 step definitions. Not all of them start with the word `the`; some of these are parameterized further, and we're already using quotes (both single and double) for their parameters. Our cucumber test suite's scenarios consist of several thousand steps, and a fair percentage of them use these expressions. 

The `shortly afterwards` is only one example of step definitions creating ambiguity in this context, and as much as we'd like to get rid of it, we can't. (It's not a matter of test assertions taking time. We need to simulate realtime in our test suite, and some things are expected to have instantaneously visible effects, while some other actions take a small deterministic amount of time. This is what this step is for.)

I hope you understand that refactoring our test suite on such a grand scale is not an option for us at the moment. I can also see that you as maintainers have a desire to keep your code simple, so I would totally understand if you'd prefer us to maintain this patch ourselves in our fork. On the other hand, if you think there's a different API that `cucumber-cpp` could expose to help us filter step matches at runtime, we'd be happy to use that instead of this patch (which is the simplest solution that solves our problems).

paoloambrosio:
It's just a thought but would a pluggable step matcher work for you? It could be something useful to support [Cucumber Expressions](https://docs.cucumber.io/cucumber-expressions/) instead of Regexes and you could implement your own.

toh-ableton:
That could be an option, though I don't know enough about Cucumber Expressions to be able to judge. From the docs you linked, I couldn't find any spec or more than 'like regexp, but better'. Do you know if there's more info available?

muggenhor:
I don't think that's what he meant. I believe his question was whether a customizable step matcher would help you solve your problem. Cucumber Expressions are just something that could also be implemented with a custom step matcher.

A custom step matcher would be constructed from some kind of match expression (i.e. what is currently a regex) and then would have to answer the question "does it match this?" for a step-string coming from a feature file. As such you should be able to use something more expressive than a regex if you wish to do so. Would that kind of customizability work as an solution for you?

paoloambrosio:
This is roughly the interface (without `const`s for readability, and probably some errors because I typed it straight into here):

```c++
struct ExpressionMatcher {
    //ExpressionMatcher(std::string expr);
    unique_ptr<ExpressionMatch> find(std::string &expression);
    std::string & str();
}

struct ExpressionMatch {
    bool matches();
    std::vector<ExpressionSubmatch> & getSubmatches();
}

struct ExpressionSubmatch {
    std::string value();
    int position();
}
```

It's a stripped down version of the current [regex expression abstraction](https://github.com/cucumber/cucumber-cpp/blob/v0.4/include/cucumber-cpp/internal/utils/Regex.hpp).

toh-ableton:
I see, I misunderstood. Yes, I think this would work perfectly for us.

How do you image this would look like in a step definition? I.e.
```c++
GIVEN /* ??? */
{
  REGEXP_PARAM(string, button_name);
}
```

PR #152, opened 2017-06-03
Fix our CMake version dependency: 3.1

muggenhor:
I.e. inside FindGMock.cmake we're using find_dependency() from the
CMakeFindDependencyMacro module. That module only got added in CMake
3.0. Furthermore we're relying on the import target Threads::Threads,
which only gets added with CMake 3.1.

So bump our requirement to what it is we really need.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.12)
+cmake_minimum_required(VERSION 3.1)
 
 project(Cucumber-Cpp)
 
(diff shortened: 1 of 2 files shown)


konserw:
What about #140 which does also that ?

muggenhor:
That PR indeed also includes this change, but it is not about the same change. That's adding a feature, this is just the bug fix and thus smaller and quicker to get in. If you're worried about merge problems: this should be auto-mergable by git.

muggenhor:
#140 looks almost good to go, so I'll close this one. I'll just make an issue report for it instead.

PR #155, opened 2017-06-03
Use clang format for style checking

### LINKED ISSUE #151: CI system coding style verification
```cucumber
Feature: Coding style verification
  In order to reduce manual labor and simple errors
  As a developer
  I want the CI system to verify that I have adhered to the coding standard
```

In short, I believe this can be accomplished by using (a recent) `clang-format` with a proper configuration file. We could use `git clang-format` to only touch lines changed in the relevant commits and then report a failure when it changed any files (`git diff -q HEAD` returns non-zero).

muggenhor:
This first adds a clang-format configuration that should match our current style as much as possible. Subsequently another Travis build gets added that uses this config to verify that all changes (when compared to the last merge from master) in a branch (including pull requests) adhere to that style. Any differences will cause that build to fail and the differences with the style, as clang-format would format it, to be shown as a highlighted unified diff.

<del>The last commit of this PR is present _only_ to demonstrate how such a failure to adhere to the style gets reported, that commit itself should _not_ be merged back into master.</del><ins>muggenhor@8b20bbeac3331c3d8ad4443d981d1e65806742fa demonstrates how such a failure looks like.</ins> That failure looks like this: https://travis-ci.org/muggenhor/cucumber-cpp/jobs/266237870

This resolves #151.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,32 @@
+---
+Language:        Cpp
+AccessModifierOffset: -4
+AlignEscapedNewlinesLeft: true
+AllowShortFunctionsOnASingleLine: Empty
+AlwaysBreakTemplateDeclarations: true
+BinPackArguments: false
+BinPackParameters: false
+BreakBeforeBinaryOperators: All
+BreakBeforeBraces: Attach
+ColumnLimit:     100
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+ForEachMacros:   [ BOOST_FOREACH ]
+IncludeCategories: 
+  # The autodetect header should always be included last
+  - Regex:           '^<cucumber-cpp/autodetect\.hpp>$'
+    Priority:        3
+  - Regex:           '^["<]cucumber-cpp/'
+    Priority:        1
+  - Regex:           '.*'
+    Priority:        2
+IndentWidth:     4
+NamespaceIndentation: None
+PenaltyBreakString: 1000
+PenaltyExcessCharacter: 10000
+PointerAlignment: Left
+SortIncludes:    false
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
+Standard:        Cpp03
+...
+
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ matrix:
     - os: linux
       compiler: gcc
       env: GMOCK_VER=1.8.0 VALGRIND_TESTS=ON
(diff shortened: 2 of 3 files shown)


konserw:
Looks good to me 👍

muggenhor:
I think we should use this as a general guideline btw. We should not reject every PR that fails the formatting check, because in some cases a strict application of the rules just doesn't produce well formatted code (i.e. sometimes manual formatting is better). So we'll just have to use sensible judgment for this.

@paoloambrosio what's your opinion on this?

paoloambrosio:
Unless I've done something stupid, currently the format is wrong for 17 files. Are we happy with the changes that `clang-format` would do? If so, I think we should add a commit to modify the existing code base according to the enforced formatting.

We should also add instructions on how to check code formatting before sending a PR (perhaps in the contributing guidelines?).

Apart from this, I think it's a great addition :+1:

muggenhor:
@paoloambrosio:
> Unless I've done something stupid, currently the format is wrong for 17 files.

Really? I get 76 files:
```console
$ clang-format-4.0 --style=file -i (^([Bb]uild*|include))/**/*.(cpp|hpp|h) include/cucumber-cpp/**/*.(cpp|hpp|h)
$ git diff --stat
 ...
 76 files changed, 1108 insertions(+), 1190 deletions(-)
```

> Are we happy with the changes that `clang-format` would do (`main.cpp` looks much worse than before)?

Most of these changes are the result of our currently inconsistent formatting in the code base. I've tried multiple variations of options and tried to find the selection that leads to the smallest amount of differences.

Most of the formatting that `clang-format` produces is good IMO. That being said, it's not perfect and cannot be, so I believe we should still exercise our own judgment: for most of the code the way `clang-format` reformats the code is probably good, but where we believe it decreases readability we should do what we consider best instead.

> If so, I think we should add a commit to modify the existing code base according to the enforced formatting.

I'm not sure that's a good idea, that would reduce the usefulness of `git blame`. Another approach is to reformat code that we touch for other changes. This is essentially what `git-clang-format` is for, it reformats only lines (statements really) that are touched since a specified commit.

> We should also add instructions on how to check code formatting before sending a PR (perhaps in the contributing guidelines?).

Agreed.

konserw:
@muggenhor what about this? Are you going to rebase and merge?
I think it would be usefull to have it merged as soon as possible. 

In general I agree that we shoul apply this only to new commits, not to rewrite history.

PR #159, opened 2017-08-12
Support taking regex captures as arguments to the step definition's function

muggenhor:
This feature depends on the availability of C++11 (or more recent). It's main purpose is to support the transformation from code such as this:
```c++
GIVEN("some (\\d+) and a \"([^\"]+)\"") {
    REGEX_PARAM(unsigned, number);
    REGEX_PARAM(std::string, string);
}
```
to use the more familiar syntax present in this code:
```c++
GIVEN("some (\\d+) and a \"([^\"]+)\"", (const unsigned number, const std::string& string)) {
}
```

Currently all captures from the regex that aren't passed as function parameters can still be extracted by means of the `REGEX_PARAM` macro.

If C++11 is unavailable then this feature is just unavailable and all existing code should continue to work as is.

In the future it may be desirable to check, at compile-time, that the amount of captures in the regex equals the amount of function parameters. I have some prototype code I've been playing with that suggests this is relatively doable. It currently depends on ["relaxed `constexpr`" support](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3652.html) (C++14), [`constexpr` lambdas](https://isocpp.org/files/papers/N4487.pdf) (C++ 17) and ["`constexpr` if"](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0292r2.html). I may be able to rewrite this prototype to rely on just the C++14 features, but that's for the future. Heck, it may even be possible to use recursion and rely only on C++11, but that's bound to produce very difficult to read code.


konserw:
Looks very good to me 👍 I'm also happy to see modern C++ features in good use.

PR #161, opened 2017-08-18
Fix crash in Qt examples

### LINKED ISSUE #62: Segmentation fault in QApplication destructor in CalcQt
Everything works until the program is shutting down and the QApplication destructor is called:

```
Program received signal SIGSEGV, Segmentation fault.
0x004eba9a in ?? () from /usr/lib/i386-linux-gnu/libQtGui.so.4
(gdb) backtrace
#0  0x004eba9a in ?? () from /usr/lib/i386-linux-gnu/libQtGui.so.4
#1  0x004ebc0c in QPixmapCache::clear() ()
   from /usr/lib/i386-linux-gnu/libQtGui.so.4
#2  0x00484a3d in ?? () from /usr/lib/i386-linux-gnu/libQtGui.so.4
#3  0x00404a9c in QApplication::~QApplication() ()
   from /usr/lib/i386-linux-gnu/libQtGui.so.4
#4  0x011f5f51 in ?? () from /lib/i386-linux-gnu/libc.so.6
#5  0x011f5fdd in exit () from /lib/i386-linux-gnu/libc.so.6
#6  0x011dc4db in __libc_start_main () from /lib/i386-linux-gnu/libc.so.6
#7  0x0808e0b9 in _start ()
```

Happens on:
- Ubuntu Linux 12.10 x64, gcc 4.7.2, Qt 4.8.3, Boost 1.53.0
- Linux Mint 13 x32, gcc 4.6.3, Qt 4.8.1

muggenhor:
This fixes #62, the crash on termination of the step definition executable for the Qt examples. First it adds testing of this to Travis though, to get CI coverage for this. Then it deduplicates the examples' source code so we don't have to fix the same problem twice.

The fix itself just revolves around ensuring that `QApplication` doesn't outlive `main()`.

PS I'm not adding tests for this to the OSX part of Travis because it doesn't work, maybe eventually they'll solve that here: travis-ci/travis-ci#7277.


konserw:
Very good PR! I haven't thought about putting both calc itself and qapplication into one wrapper, what lead me to errors. I feel a little bit stupid to review your work especially as I don't have any comments on the code per se, so you may want to ask @paoloambrosio for anther review; but I would be very happy to exercise my knew commit power and merge this :)

I've got 3 thoughts about gui testing you've added:
* I've also thought about xvfb but my experiments weren't as neat as this. 
* Have you thought about displaylink test results from the examples in the same way as from features target? I think it is --format progress flag
* It's a pity that it doesn't work on osx, but I think it should be doable on appveyor windows build - have you tried?

muggenhor:
@konserw:
> Very good PR!

Thanks :-)

> I haven't thought about putting both calc itself and qapplication into one wrapper, what lead me to errors. I feel a little bit stupid to review your work especially as I don't have any comments on the code per se

I would have merged this right away, but thought it to be good to have another pair of eyeballs on this. Considering that you've looked at the exact same code for the same purpose, your insight does matter. Not having any comments isn't a bad thing btw.

> but I would be very happy to exercise my knew commit power and merge this :)

Before doing something like that I think it's a good thing for us to merge #146 and you to read the "maintainer's guide" section of `CONTRIBUTING.md` in there.

> * Have you thought about displaylink test results from the examples in the same way as from features target? I think it is --format progress flag

I've thought about it, but I actually like the fully verbose output (aka `--format pretty`) better in CI. Especially when something goes wrong. It doesn't just give you a single red character when failing, instead it gives you the scenario and the location where it failed.

> * It's a pity that it doesn't work on osx, but I think it should be doable on appveyor windows build - have you tried?

I've tried, and failed, mostly because I cannot seem to manage to start background processes on Windows' two shells (cmd and power shell). Nor do I have a Windows development machine available to experiment with. So just like with OSX I've decided not to allow for "the perfect is the enemy of the good" to apply here and just put in the parts that do work.

PR #162, opened 2017-08-19
Track test coverage using coveralls service

### LINKED ISSUE #123: Code Coverage?

konserw:
## Summary
Track test coverage using coveralls service
Url: https://coveralls.io
Implementation using lcov and gem coveralls-lcov

## Details
After we turned down CMake-based implementation, and I've failed to get python script from here: https://github.com/eddyxu/cpp-coveralls to work; I've based this implementation on this post: https://gronlier.fr/blog/2015/01/adding-code-coverage-to-your-c-project/ which is using lcov and ruby gem just to upload results.

## Motivation and Context

This fixes #123

## How Has This Been Tested?

No tests needed.

## Types of changes

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: cpp
-os:
-  - linux
+os: linux
 sudo: required
 dist: trusty
 compiler:
@@ -12,6 +11,9 @@ env:
   - GMOCK_PATH=/usr/src/gmock #1.6.0 from ubuntu trusty repo
 matrix:
   include:
+    - os: linux
+      compiler: gcc
+      env: GMOCK_VER=1.8.0 COVERALLS_SERVICE_NAME=travis-ci
     - os: linux
       compiler: gcc
       env: GMOCK_VER=1.8.0 VALGRIND_TESTS=ON
@@ -38,9 +40,12 @@ addons:
       - qtbase5-dev
       - xvfb
 
-before_install:
+install:
   - if [[ "${TRAVIS_OS_NAME}" = "osx" ]]; then brew update && brew install ninja qt5; fi
+  - if [ -n "${COVERALLS_SERVICE_NAME}" ]; then gem install coveralls-lcov; fi
+  - if [ -n "${COVERALLS_SERVICE_NAME}" ]; then wget http://ftp.de.debian.org/debian/pool/main/l/lcov/lcov_1.13.orig.tar.gz; tar xf lcov_1.13.orig.tar.gz; sudo make -C lcov-1.13/ install; fi
+# we have to build lcov on our own as version 1.10 in apt is incompatible with coveralls gem; It is not possible to install lcov from deb as it doesn't recognize installed perl package.
 
-script: ./travis.sh 
+script: ./travis.sh
 
 # default notifications: https://docs.travis-ci.com/user/notifications#Notifications
--- a/travis.sh
+++ b/travis.sh
@@ -20,6 +20,8 @@ cmake -E chdir build cmake \
     -G Ninja \
     -DCUKE_ENABLE_EXAMPLES=on \
     ${CMAKE_PREFIX_PATH:+"-DCMAKE_PREFIX_PATH=${CMAKE_PREFIX_PATH}"} \
+    ${COVERALLS_SERVICE_NAME:+"-DCMAKE_BUILD_TYPE=Debug"} \
+    ${COVERALLS_SERVICE_NAME:+"-DCMAKE_CXX_FLAGS='--coverage'"} \
     ${VALGRIND_TESTS:+"-DVALGRIND_TESTS=${VALGRIND_TESTS}"} \
     ${GMOCK_PATH:-"-DGMOCK_VER=${GMOCK_VER}"} \
     ${GMOCK_PATH:+"-DGMOCK_SRC_DIR=${GMOCK_PATH}"} \
@@ -78,4 +80,14 @@ if [ -n "${XVFBPID:-}" ]; then
     # Stop virtual X display server
     kill $XVFBPID
     wait
+fi
+
+if [ -n "${COVERALLS_SERVICE_NAME}" ]; then
+    lcov --directory ./build --capture  --output-file coverage.info
+    lcov --remove coverage.info "/usr/*" --output-file coverage.info
+    lcov --remove coverage.info "${TRAVIS_BUILD_DIR}/build/*" --output-file coverage.info
+    lcov --remove coverage.info "${TRAVIS_BUILD_DIR}/tests/*" --output-file coverage.info
+    lcov --remove coverage.info "${TRAVIS_BUILD_DIR}/include/json_spirit/*" --output-file coverage.info
+    lcov --list coverage.info
+    coveralls-lcov coverage.info
 fi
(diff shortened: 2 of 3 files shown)


HISTORY.md
```diff
@@ -5,6 +5,7 @@
 * Listen on localhost by default to avoid firewall warnings ([#158](https://github.com/cucumber/cucumber-cpp/pull/158) Nik Reiman)
 * Better integrate Qt into buildsystem, it can now be disabled, and test it in CI ([#160](https://github.com/cucumber/cucumber-cpp/pull/160) Kamil Strzempowicz & Giel van Schijndel)
 * Support taking regex captures as arguments to the step definition's function ([#159](https://github.com/cucumber/cucumber-cpp/pull/159) Giel van Schijndel)
+* Track test coverage using coveralls service ([#162](https://github.com/cucumber/cucumber-cpp/pull/162) Kamil Strzempowicz)
```
muggenhor: This is not a feature of `cucumber-cpp`. It's not (directly) useful for the users of this library. Instead it's useful for us, the developers, as such I believe this doesn't need to be mentioned in this file and definitely not in the "new features" section.

konserw: ok, i'll remove that.

konserw:
I don't know why travis job for coverage is red. my coverage script exits with 0, but build exits with 1?

konserw:
Ok, I've found it - now it should be all green

konserw:
Ad meritum: Should we include in coverage report:
* tests itself (now excluded)?
* qmock (now excluded)?
* json spirit (now included)?

cover.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
+set -e #break script on non-zero exitcode from any command
+set -x #display command being executed
+CTEST_OUTPUT_ON_FAILURE=ON
+export CTEST_OUTPUT_ON_FAILURE
+
+wget http://ftp.de.debian.org/debian/pool/main/l/lcov/lcov_1.11.orig.tar.gz
```
muggenhor: Why don't you use `lcov` from APT?

cover.sh
```diff
@@ -0,0 +1,30 @@
...
+tar xf lcov_1.11.orig.tar.gz
+sudo make -C lcov-1.11/ install
+gem install coveralls-lcov
+
+cmake -E make_directory build
+# init coverage to 0 (optional)
+lcov --directory ./build --zerocounters
+cmake -E chdir build cmake \
+    -G Ninja \
+    -DCUKE_ENABLE_EXAMPLES=off \
```
muggenhor: Why not? These are run during CI, so are in effect tests.

cover.sh
```diff
@@ -0,0 +1,30 @@
...
+    -DCMAKE_BUILD_TYPE=Debug \
+    -DCMAKE_CXX_FLAGS='--coverage' \
+    ${GMOCK_PATH:-"-DGMOCK_VER=${GMOCK_VER}"} \
+    ..
+cmake --build build
+cmake --build build --target test
+
```
muggenhor: Why don't you run the feature tests? They provide coverage too.

.travis.yml
```diff
@@ -38,9 +40,9 @@ addons:
       - qtbase5-dev
       - xvfb
 
-before_install:
+install:
   - if [[ "${TRAVIS_OS_NAME}" = "osx" ]]; then brew update && brew install ninja qt5; fi
 
-script: ./travis.sh 
+script: if [ -z "$COVERALLS_SERVICE_NAME" ]; then ./travis.sh; else ./cover.sh; fi
```
muggenhor: Why the separate file? It looks to me like the only necessary difference is that of installing and running lcov followed by submission of the coverage results, that should fit in a single conditional block quite well.

cover.sh
konserw: ok

cover.sh
konserw: as above

cover.sh
konserw: Blog post I linked mentioned that lcov from apt is old and didn't work fo him. I can check how it would work for us.

.travis.yml
konserw: ok, I'll merge them

konserw:
with lcov from repo it failed: https://travis-ci.org/cucumber/cucumber-cpp/jobs/266579176 - let's see if installing from deb would work out better.

muggenhor:
I think your attempt in cab194daf808f8d48f9241426692c9d513479486 would work if you add `perl` to the package list, seems the absence of that is the only reason it failed.

konserw:
I've noticed this perl thing, but thought it wasn't worth installing perl just to get deb working. We can try and see which is faster.

muggenhor:
It seems the syntax for specifying dependencies on whatever architecture (`any`) has changed, so installing `lcov` from something more recent than trusty in trusty is never going to work if it has an `any` dependency.

konserw:
indeed - all in all I think building lcov is not so bad after all :)

muggenhor:
Agreed, but could you add a comment to the script explaining that we're building it to get lcov >= 1.11, and that just grabbing the .deb from more recent versions doesn't work?

konserw:
@muggenhor of course

konserw:
For the record: to use more recent version of lcov we would need to change relative paths into full paths in the remove parameter.
As for failed appveyor build - I think it was problem with github. Retriggering build should fix it

konserw:
Squashed again - now it should be ready to merge

travis.sh
```diff
@@ -79,3 +81,14 @@ if [ -n "${XVFBPID:-}" ]; then
     kill $XVFBPID
     wait
 fi
+
+if [ -n "${COVERALLS_SERVICE_NAME}" ]; then
+echo "travis build dir: ${TRAVIS_BUILD_DIR}"
```
muggenhor: indentation

travis.sh
```diff
@@ -79,3 +81,14 @@ if [ -n "${XVFBPID:-}" ]; then
...
+    lcov --directory ./build --capture  --output-file coverage.info
+    lcov --remove coverage.info "/usr/*" --output-file coverage.info
+    lcov --remove coverage.info "${TRAVIS_BUILD_DIR}/build/gmock/*" --output-file coverage.info
```
muggenhor: I think you can exclude `build/*`. That would also get rid of the generated `moc_CalculatorWidget.cpp` and `cpp_steps.cpp` neither of which we care about for coverage.

travis.sh
konserw: sorry - I should have removed it.

travis.sh
konserw: good catch - I haven't noticed that ;)

muggenhor:
Oh, just noticed this: could you remove the features and examples from coverage tracking too? Other than that this looks ready to merge. :+1:

muggenhor:
This looks good to merge. In this case I don't think you need to wait on Travis' OSX build (nothing changed that can affect OSX).

konserw:
Ok, but do I need to rebase onto master once again as it has been changed in the meantime?

muggenhor:
I'm assuming you did something like this.

```console
$ git merge --no-ff --no-commit konserw/rebasedGemCoverage && git commit
```

The end result looks good at least.

konserw:
Not exactly, but I'm glad that end result is OK ;)

PR #164, opened 2017-08-23
Fix issue #81 by intercepting messages from BOOST_*_MESSAGE macros

### LINKED ISSUE #81: BOOST_PREDICATE_MESSAGE(boolean, string message)
When using boost testing tools the messages do not show up when using the above evaluation. According to boost, if the predicate fails the custom message should appear with the failure. Instead I only receive a general message indicating failure with no string message attached. Is this a supported feature?

konserw:
## Summary

Intercept messages from BOOST_*_MESSAGE macros

## Motivation and Context

Fixes #81

## How Has This Been Tested?

It has been tested manually by adding following line to BoostCalc example and running that example:
`BOOST_CHECK_MESSAGE(false, "Message");`

## Types of changes

## Checklist:


src/drivers/BoostDriver.cpp
```diff
@@ -60,8 +60,8 @@ class CukeBoostLogInterceptor : public ::boost::unit_test::unit_test_log_formatt
     void log_exception_finish( std::ostream& ) {};
 
     void log_entry_start( std::ostream&, log_entry_data const&, log_entry_types /*let*/) {};
-    void log_entry_value( std::ostream&, const_string /*value*/);
-    void log_entry_value( std::ostream&, lazy_ostream const& /*value*/) {};
+    void log_entry_value( std::ostream&, const_string value);
+    void log_entry_value( std::ostream&, lazy_ostream const& value);
```
muggenhor: Please don't do this: this causes warnings about "unused parameters", that's the reason it was commented out like that in the first place.

konserw: But we are actually using them?

konserw: grep for unused and for param hasn't show me any warning

muggenhor: Oh, nevermind, I didn't notice that you changed the empty inline function (`{}`) to just a declaration with the function's definition elsewhere.

As for warnings: you should compile with `-Wall -Wextra` to see any. I think we should probably consider building with that one plus `-Werror` by default on our CI systems at least.

The old code (with empty inline function definition) without commented out parameter produces this with `-DCMAKE_CXX_FLAGS='-Wall -Wextra -Werror':
```
../src/drivers/BoostDriver.cpp:64:62: error: unused parameter 'value' [-Werror,-Wunused-parameter]                                                                 void log_entry_value( std::ostream&, lazy_ostream const& value) {};                                                               
                                                             ^
```

konserw: I thought we are doing so for debug builds, but now I can see that there is only -Weffc++ being set for those. We could add those flags to CMAKE_CXX_FLAGS_DEBUG in CMakelists ?

muggenhor: I've just submitted 3ec9d0367b73abfeb0b05eebf612f551e88d9dad which enables warnings for all GCC/Clang builds. I don't think it makes sense to make it dependent on the build type. Adding `-Werror` is something we should do only when the warnings are fixed.

muggenhor:
Can you also add a test (perhaps in `tests/integration/drivers`) that fails without this change and succeeds with it?

konserw:
Gladly, but I'm not sure how

konserw:
For failing test see https://travis-ci.org/cucumber/cucumber-cpp/jobs/268059750

konserw:
@muggenhor - could you take a look at the test I've added? I'm not good with ruby... Also maybe we should rethink how the "support code" is inserted to every feature with C++ code?

features/specific/predicate_message.feature
```diff
@@ -0,0 +1,70 @@
+@wip
+Feature: Predicate Message Feature
+
+  This is just a simple feature meant to test the fix for issue #81.
+  Problem was Cucumber-Cpp didn't recoginize failed asserations
+  when using following Boost macros:
```
muggenhor: Rephrase without the issue please. I.e. something like "Support Boost.Test's different validation and reporting macros..."

features/specific/predicate_message.feature
```diff
@@ -0,0 +1,70 @@
+@wip
```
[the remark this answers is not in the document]
konserw: or replace with @boost ?

konserw:
For failing test see https://travis-ci.org/cucumber/cucumber-cpp/jobs/268059750

CMakeLists.txt
```diff
@@ -216,7 +216,7 @@ else()
         add_executable(e2e-steps EXCLUDE_FROM_ALL ${CUKE_DYNAMIC_CPP_STEPS})
         # Mark this file as generated so it isn't required at CMake generation time (it is necessary when the target gets built though)
         set_source_files_properties(${CUKE_DYNAMIC_CPP_STEPS} PROPERTIES GENERATED TRUE)
-        target_link_libraries(e2e-steps ${CUKE_LIBRARIES})
+        target_link_libraries(e2e-steps ${CUKE_LIBRARIES} ${Boost_UNIT_TEST_FRAMEWORK_LIBRARY})
```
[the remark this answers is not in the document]
konserw: Good point.

features/specific/predicate_message.feature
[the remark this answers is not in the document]
muggenhor: Yes, that's a good one

features/step_definitions/cucumber_cpp_mappings.rb
```diff
@@ -228,6 +228,10 @@ def append_step_definition(step_name, code) # todo params parameter?
 EOF
   end
 
+  def no_support_code(code)
```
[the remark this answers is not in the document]
konserw: this is without the support code (code from support folder)

features/specific/predicate_message.feature
konserw: done

konserw:
@muggenhor how do you like it now? I kinda changed it other way around...

konserw:
I've pasted wrong link twice. For failing test see this one: https://travis-ci.org/konserw/cucumber-cpp/jobs/268053073

CMakeLists.txt
```diff
@@ -248,9 +254,9 @@ else()
                 )
         endfunction(add_feature_target)
 
-        add_feature_target(features --format progress)
-        add_feature_target(features-pretty --format pretty)
-        add_feature_target(features-wip --format pretty --tags @wip)
+        add_feature_target(features         ${CUKE_E2E_TAGS} --format progress)
```
muggenhor: Just add it before `${ARGN}` in the `add_feature_target` function instead. That reduces repetition.

features/specific/predicate_message.feature
```diff
@@ -0,0 +1,72 @@
+@boost
+Feature: Predicate Message Feature
```
paoloambrosio: Either this should be in a `boost` directory or it should specify that it is related to Boost. Please remove the redundant `Feature` at the end.

features/specific/predicate_message.feature
```diff
@@ -0,0 +1,72 @@
...
+
+  This is just a simple feature meant to test boost's 
+  validation macros that also output custom message.
+  Problem was Cucumber-Cpp didn't capture those messages
+  and even didn't recognize failed asserations.
```
paoloambrosio: The description should capture the behaviour, not why you did it (you should find that in the PR).

features/specific/predicate_message.feature
```diff
@@ -0,0 +1,72 @@
...
+  Expected results when using following Boost macros:
+    BOOST_WARN_MESSAGE( predicate, message ) -> should pass
+    BOOST_CHECK_MESSAGE( predicate, message ) -> should fail
+    BOOST_REQUIRE_MESSAGE( predicate, message ) -> should fail
```
paoloambrosio: Redundant and can get out of sync with the scenarios.

features/specific/predicate_message.feature
```diff
@@ -0,0 +1,72 @@
...
+
+  Scenario: BOOST_WARN_MESSAGE Scenario
```
paoloambrosio: Please remove redundant `Scenario` at the end.

features/specific/predicate_message.feature
```diff
@@ -0,0 +1,72 @@
...
+      """
+      #include <boost/test/unit_test.hpp>
+      #include <cucumber-cpp/autodetect.hpp>
+
+      GIVEN("a step") {
+        BOOST_REQUIRE_MESSAGE(false, "boost message");
+      }
+      """
+    When Cucumber runs the feature
+    Then the failure message "boost message" is output
```
paoloambrosio: Shall we condense all those scenarios into one scenario outline?

```
  Scenario Outline: <check>
    Given the following feature:
      """
      Feature: Feature name
        Scenario: Scenario name
          Given a step
      """
    And a step definition file with:
      """
      #include <boost/test/unit_test.hpp>
      #include <cucumber-cpp/autodetect.hpp>
      GIVEN("a step") {
        <check>(false, "boost message");
      }
      """
    When Cucumber runs the feature
    Then the scenario <passes or fails?>

    Examples:
      | check                 | passes or fails?                   |
      | BOOST_WARN_MESSAGE    | passes                             |
      | BOOST_CHECK_MESSAGE   | fails with message "boost message" |
      | BOOST_REQUIRE_MESSAGE | fails with message "boost message" |
```

konserw: Neat 👍 I was thinking about something like that - I just haven't used outlines before...

konserw:
I've created new feature file based on @paoloambrosio suggestions - now it contains more boost's macros.
Now I've also rebased against newest master.

Build is failing due to #169

paoloambrosio:
After adding the other scenarios, there is again some duplication. I've tried experimenting with a single `ScenarioOutline` with all the examples, and I think it would be very readable:
```
    Examples:
      | check                                         | passes or fails?                    |
      | BOOST_CHECK(false)                            | fails                               |
      | BOOST_CHECK_MESSAGE(false, "boost message")   | fails with message "boost message"  |
      | BOOST_ERROR("boost message")                  | fails with message "boost message"  |
      | BOOST_FAIL("boost message")                   | fails with message "boost message"  |
      | BOOST_REQUIRE(false)                          | fails                               |
      | BOOST_REQUIRE_MESSAGE(false, "boost message") | fails with message "boost message"  |
      | BOOST_WARN(false)                             | passes                              |
      | BOOST_WARN_MESSAGE(false, "boost message")    | passes                              |
```

What do you think?

konserw:
Looks good - I've commited this change.

konserw:
All green - I just had to rebase on top of master after change in travis scripts. I'll merge now.

PR #166, opened 2017-08-24
CHANGELOG update for cucumber/cucumber #251

jaysonesmith:
## Summary

CHANGELOG.md updates as per [cucumber/cucumber #251](https://github.com/cucumber/cucumber/issues/251)

## Details

Nothing huge in this change. Changed the file's name, added the contributing link to the top, and added my change to the list.

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [x] Documentation update

PR #167, opened 2017-08-24
Revert "CHANGELOG update for cucumber/cucumber #251"

jaysonesmith:
Reverts cucumber/cucumber-cpp#166

Sorry, @paoloambrosio!


paoloambrosio:
I don't think there was anything wrong with the previous change, apart from merging the PR without approval and before checks have finished running.

@muggenhor @konserw I have now enabled master branch protection to prevent that from happening in the future.

I will close this and leave the other one merged. @

Issue #169, opened 2017-09-13
BoostCalcQt segfaults on travis


konserw:
## Summary

Travis fails for all (3 of them) PRs: https://travis-ci.org/cucumber/cucumber-cpp/pull_requests since 2 weeks ago
Reason is boost version of calcqt example, not sure why as it runs well on my linux machine.

## Possible Solution

I'll check if it fails also when using unix sockets, but it is more like workaround, not the solution.

## Exact error message:
```

+wait %

+[ -f build/examples/CalcQt/QtTestCalculatorQtSteps -a -n :99 ]

+[ -f build/examples/CalcQt/BoostCalculatorQtSteps -a -n :99 ]

+sleep 1

+build/examples/CalcQt/BoostCalculatorQtSteps

+cucumber examples/CalcQt

# language: en

Feature: Addition

  In order to avoid silly mistakes

  As a math idiot 

  I want to be told the sum of two numbers

  Scenario Outline: Add two numbers       # examples/CalcQt/features/addition.feature:7

    Given I just turned on the calculator # examples/CalcQt/features/addition.feature:8

    When I press <button1>                # examples/CalcQt/features/addition.feature:9

    And I press add                       # examples/CalcQt/features/addition.feature:10

    And I press <button2>                 # examples/CalcQt/features/addition.feature:11

    And I press calculate                 # examples/CalcQt/features/addition.feature:12

    Then the display should show <result> # examples/CalcQt/features/addition.feature:13

    Examples: 

      | button1 | button2 | result |

      | 2       | 3       | 5      |

      Timed out calling wire server with message 'invoke' (Timeout::Error)

      examples/CalcQt/features/addition.feature:17:in `Given I just turned on the calculator'

      examples/CalcQt/features/addition.feature:8:in `Given I just turned on the calculator'

Timed out calling wire server with message 'step_matches' (Timeout::Error)

/home/travis/.rvm/gems/ruby-2.4.1/gems/cucumber-2.0.0/lib/cucumber/wire_support/connection.rb:43:in `block in fetch_data_from_socket'

/home/travis/.rvm/rubies/ruby-2.4.1/lib/ruby/2.4.0/timeout.rb:108:in `timeout'

/home/travis/.rvm/gems/ruby-2.4.1/gems/cucumber-2.0.0/lib/cucumber/wire_support/connection.rb:43:in `fetch_data_from_socket'

[...]

PR #170, opened 2017-09-18
Switch to old travis image and enable Xvfb tests on OSX

konserw:
It is more like workaround for issue #169 - using old travis image fixes our false fails in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+group: deprecated-2017Q3
+
 language: cpp
 os: linux
 sudo: required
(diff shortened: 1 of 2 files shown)


konserw:
I've also added running Xvfb examples on OSX and moved Xvfb -related code to functions. 
I would merge those changes in order to proceed with development process while in the mean time proper resolution to this issue may appear on travis or boost side

paoloambrosio:
Fixing an issue we cannot reproduce in local is going to be tough. I am in favour of unblocking the CI build to merge pending PRs.

@muggenhor what do you think?

muggenhor:
Yes, merging this is probably the best, at least on the short term. I'm only worried that on Travis' side they'll remove the pinned image at some point in the future. But we can worry about that when it happens.

PR #173, opened 2017-10-12
Lost changelog line

konserw:
Lost changelog line.
Unfortunatelly github hook stopped both amending merge commit and commiting separate commit without PR...


konserw:
@paoloambrosio sure thing. I've even edited CHANGELOG, just forgot to stage it for commit...

Issue #190, opened 2018-05-19
QtTestDriverTest doesn't get built nor executed and leaks


muggenhor:
Because of a typo in `tests/CMakeLists.txt` QtTestDriverTest doesn't get built and thus not executed either:
```cmake
if(Qt5TEST_FOUND)
```
This should be (note the title case instead of full upper case):
```cmake
if(Qt5Test_FOUND)
```

When fixing that however, I'm getting a memory leak: https://travis-ci.org/muggenhor/cucumber-cpp/jobs/381075041

@konserw Given that you originally introduced the QtTest driver in #165, can you have a look at this and try to fix it?

This is the output:
```
108/109 Test #108: QtTestDriverTest .................................................................***Failed    1.26 sec
==10192== Memcheck, a memory error detector
==10192== Copyright (C) 2002-2013, and GNU GPL'd, by Julian Seward et al.
==10192== Using Valgrind-3.10.1 and LibVEX; rerun with -h for copyright info
==10192== Command: /home/travis/build/muggenhor/cucumber-cpp/build/tests/QtTestDriverTest
==10192== 
SUCCESS (The test body has not been run)
SUCCESS (The test body has been run)
= Step invocation =
SUCCESS (Succeeding step)
SUCCESS (Failing step)
SUCCESS (Failing step has a message)
SUCCESS (Pending step with no description - result)
SUCCESS (Pending step with no description - description)
SUCCESS (Pending step with description - result)
SUCCESS (Pending step with description - description)
SUCCESS (Inexistent step)
= Constructors and destructors =
== Succeeding step ==
SUCCESS (Contexts created after invoke)
SUCCESS (Contexts destroyed after invoke)
SUCCESS (Contexts created after end scenario)
SUCCESS (Contexts destroyed after end scenario)
== Failing step ==
SUCCESS (Contexts created after invoke)
SUCCESS (Contexts destroyed after invoke)
SUCCESS (Contexts created after end scenario)
SUCCESS (Contexts destroyed after end scenario)
= Step failure description is reset =
SUCCESS (Failing step description is set)
SUCCESS (Failing step description is the same)
==10192== 
==10192== HEAP SUMMARY:
==10192==     in use at exit: 116 bytes in 3 blocks
==10192==   total heap usage: 734 allocs, 731 frees, 68,470 bytes allocated
==10192== 
==10192== 4 bytes in 1 blocks are definitely lost in loss record 1 of 3
[...]

PR #191, opened 2018-05-19
Use dependencies via target_link_libraries exclusively

muggenhor:
This allows all dependencies to be resolved in CMake's generation step
instead of before that time during script execution.

Additionally properties such as transitive dependencies are now handled
properly and include directories of dependencies automatically added to
the include path.

NOTE: This includes a workaround for cucumber/cucumber-cpp#190 which was
      originally discovered when doing this change.

This makes #147 work properly when used fully on our CI setups (AppVeyor + Travis).


muggenhor:
FYI: the coverage decrease is because the total line count increased more than the covered line count. Covered lines on master 969/1002 -> covered lines by this PR 991/1025.

I.e. the percentage is a misleading metric

PR #208, opened 2018-12-21
Fix #203: QtTestCalculatorQtSteps is not generated because of a typo in CMakeLists.txt

### LINKED ISSUE #203: examples/CalcQt/QtTestCalculatorQtSteps is not built
## Summary
<build>/examples/CalcQt/QtTestCalculatorQtSteps is missing after build. 

## Possible Solution
In examples.CalcQt/CMakeLists.txt:
if(Qt::Test)
Should be changed to 
if(TARGET Qt::Test)

## Environment
Ubuntu 18.04
Qt 5.11.2
cmake 3.10.2

sergey-bon:
## Summary

Fix a typo that prevents `build/examples/CalcQt/QtTestCalculatorQtSteps` from being generated.  
It is, however, leaking memory. See #212 

## How Has This Been Tested?

Clone and build with examples, look for `build/examples/CalcQt/QtTestCalculatorQtSteps`. It is missing. Apply this change, rebuild, it appears.

## Types of changes

## Checklist:

### DIFF
--- a/examples/CalcQt/CMakeLists.txt
+++ b/examples/CalcQt/CMakeLists.txt
@@ -15,10 +15,8 @@ if(TARGET Qt::Core AND TARGET Qt::Gui AND TARGET Qt::Widgets AND TARGET Qt::Test
     add_executable(calcqt src/CalcQt.cpp)
     target_link_libraries(calcqt PRIVATE libcalcqt Qt::Widgets)
 
-    if(Qt::Test)
-        add_executable(QtTestCalculatorQtSteps features/step_definitions/QtTestCalculatorQtSteps)
-        target_link_libraries(QtTestCalculatorQtSteps PRIVATE libcalcqt Qt::Test cucumber-cpp)
-    endif()
+    add_executable(QtTestCalculatorQtSteps features/step_definitions/QtTestCalculatorQtSteps)
+    target_link_libraries(QtTestCalculatorQtSteps PRIVATE libcalcqt Qt::Test cucumber-cpp)
 
     if(TARGET Boost::unit_test_framework)
         add_executable(BoostCalculatorQtSteps features/step_definitions/BoostCalculatorQtSteps.cpp)
--- a/travis.sh
+++ b/travis.sh
@@ -70,7 +70,9 @@ cmake --build build
 cmake --build build --target test
 cmake --build build --target features
 
-startXvfb # Start virtual X display server
+#
+# Execute Calc examples
+#
 
 for TEST in \
     build/examples/Calc/GTestCalculatorSteps \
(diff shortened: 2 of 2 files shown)


examples/CalcQt/CMakeLists.txt
```diff
@@ -15,7 +15,9 @@ if(TARGET Qt::Core AND TARGET Qt::Gui AND TARGET Qt::Widgets AND TARGET Qt::Test
     add_executable(calcqt src/CalcQt.cpp)
     target_link_libraries(calcqt PRIVATE libcalcqt Qt::Widgets)
 
-    if(Qt::Test)
+    if(TARGET Qt::Test)
```
paoloambrosio: Shouldn't we remove `AND TARGET Qt::Test` in the outer if statement?

examples/CalcQt/CMakeLists.txt
```diff
@@ -15,7 +15,9 @@ if(TARGET Qt::Core AND TARGET Qt::Gui AND TARGET Qt::Widgets AND TARGET Qt::Test
...
+        message (WARNING "QtTestCalculatorQtSteps is leaking (see issue #212)")
```
paoloambrosio: Don't think this warning belongs to the build file. At this point we also don't know if the memory leak is in `QtTestCalculatorQtSteps` or in QtTest support.

examples/CalcQt/CMakeLists.txt
sergey-bon: You're right.

examples/CalcQt/CMakeLists.txt
sergey-bon: Warning removed.

examples/CalcQt/CMakeLists.txt
paoloambrosio: That is probably correct. They all depend on QTest, but only one uses it as a test runner.

sergey-bon: I wasn't sure if you resolved this 1 minute after my force push or it's a glitch so I unresolved this back again.

paoloambrosio: I did resolve it tight after you pushed the commit, waiting for the build, as it did address my concerns :-)

sergey-bon:
@paoloambrosio is it okay to merge?

paoloambrosio:
Pretty sure the build is going to fail unless we merge #218 first. I'll try and merge a few PRs later today.

paoloambrosio:
Should we remove `-f "${TEST}" -a` in [`travis.sh`](https://github.com/cucumber/cucumber-cpp/blob/master/travis.sh#L94) to have the build fail if the file is not created?

paoloambrosio:
Can you rebase this to my [`verify-pr-208` branch](https://github.com/paoloambrosio/cucumber-cpp/tree/verify-pr-208)?

It makes the build [fail](https://travis-ci.org/paoloambrosio/cucumber-cpp/builds/479757536) unless [your patch is applied](https://travis-ci.org/paoloambrosio/cucumber-cpp/builds/479759895).

sergey-bon:
@paoloambrosio done

paoloambrosio:
Something went wrong in the rebase. It should look something like this branch: https://github.com/paoloambrosio/cucumber-cpp/commits/sergey-bon-patch-1-rebased

From current master it should add my commit `91033b8` and your `e4700dc`. This contains also other three commits.

sergey-bon:
That's because I rebased on both verify-pr-208 and master.

I've pulled master, re-applied fix, and rebased on verify-pr-208.
It should be fine now.

PR #223, opened 2019-06-15
Add ability to send embeddings through the wire protocol

Sigill:
## Motivation and Context

I'm looking into cucumber-cpp for my company.
We need to be able to embed text or images into the reports, but that feature is not available when using the wire protocol.

This patch is the 4th of a series of 4, respectively impacting _cucumber-ruby_, _cucumber-ruby-core_, _cucumber-ruby-wire_ and _cucumber-cpp_.

Basically, the idea is to allow embeddings to be specified in the success or fail responses sent through the wire protocol. E.g.:
```
["success",{"embeddings":[{"label":"Embedded text","mime_type":"text/plain","src":"Some text"}]}]
["fail",{"embeddings":[{"label":"Embedded image","mime_type":"image/png;base64","src":"iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5ErkJggg=="}],"message":"There was no  banana"}]
```

The patch on _cucumber-ruby_ (cucumber/cucumber-ruby#1354) will let formatters access embeddings stored in a step result.
The patch on _cucumber-ruby-core_ (cucumber/cucumber-ruby-core#170) will store in the step result the embeddings returned by the step invocation.
The patch on _cucumber-ruby-wire_ (cucumber/cucumber-ruby-wire#20) will allow the embeddings stored in the wire response to be returned by the wire-based step invocation process.
The patch on _cucumber-cpp_ (cucumber/cucumber-cpp#223) will allow C++ steps to send embeddings through wire responses.

Note: embeddings were not considered to be useful for pending steps, but support should be possible if required.

## Details

This patch:
- Redesign the `invokeStep()` method to return an `InvokeResult` instead of relying on implicit success or exceptions.
- Add an `embed()` function to allow steps to declare embeddings. The embeddings of a step are temporarily stored within that step.
- Once the result of a step is known, store the embeddings in the `InvokeResult`, then in the `WireResponse` object.
- Serialize the embeddings from the `WireResponse` in the packet to be sent.

Patches on _cucumber-ruby_, _cucumber-ruby-core_, _cucumber-ruby-wire_ have been submitted for integration in cucumber 4.x, but the changes on _cucumber-cpp_ should be backward compatible with cucumber 2.x. Embeddings will just be ignored.

## How Has This Been Tested?

Unit test on the redesign of the `invokeStep()` method, serialization of embeddings and creation of `WireResponse` from `InvokeResult`.
Still need to look at E2E tests (suggestions welcomed).

## Types of changes

## Checklist:

- [x] I've added tests for my code.

### DIFF
--- a/examples/Calc/features/step_definitions/CalculatorSteps.cpp
+++ b/examples/Calc/features/step_definitions/CalculatorSteps.cpp
@@ -25,12 +25,19 @@ WHEN("^I press add") {
 WHEN("^I press divide") {
     ScenarioScope<CalcCtx> context;
 
+    embed("Some text", "text/plain", "Embedded text");
+
     context->result = context->calc.divide();
 }
 
 THEN("^the result should be (.*) on the screen$") {
     REGEX_PARAM(double, expected);
     ScenarioScope<CalcCtx> context;
+
+    embed("iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAYAAAAfFcSJAAAADUlEQVR42mP8z8BQDwAEhQGAhKmMIQAAAABJRU5E"
+          "rkJggg==",
+          "image/png;base64",
+          "Embedded image");
 
     EXPECT_EQ(expected, context->result);
 }
--- a/include/cucumber-cpp/internal/CukeEngine.hpp
+++ b/include/cucumber-cpp/internal/CukeEngine.hpp
@@ -8,6 +8,7 @@
 #include <boost/multi_array.hpp>
 
 #include <cucumber-cpp/internal/CukeExport.hpp>
+#include <cucumber-cpp/internal/step/StepManager.hpp>
 
 namespace cucumber {
 namespace internal {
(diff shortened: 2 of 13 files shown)


jermus67:
This PR is closed without merging.

The PR was the last of sequence of four PRs, that had to be merged in a specific sequence as each PR in the series depended on the previous on having been merged before.

* The first PR was to update the json parser in cucumber-ruby, this PR has been merged.
* The second PR was to update cucumber-ruby-core to allow embeddings in step results, this one never got merger (closed by stale-bot).
* The third PR was to update cucumber-ruby-wire to support these embeddings over the wire protocol (also close by stale-bot).

Since the second and third PR have been closed without merging, there is no point in merging this one. If this functionality is required first the two other PRs need to be revived before it makes sense to have another look at this one.

PR #227, opened 2019-12-07
Fix API change in boost::unit_test::unit_test_log_formatter

LukasWoodtli:
## Summary
There is an API change in `unit_test_log_formatter` from Boost.Test.
## Details

`boost::unit_test::unit_test_log_formatter::log_build_info` now expects a second argument.

## Motivation and Context

There is now a preprocessor define to support building also with Boost.Test from 1.70.

## How Has This Been Tested?

Manual build with Boost 1.71. There's no package for Boost >= for trusty (as used on travis).

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue).

## Checklist:

- [x] It is my own work, its copyright is implicitly assigned to the project and no substantial part of it has been copied from other sources (including [Stack Overflow](https://stackoverflow.com/)). In rare occasions this is acceptable, like in CMake modules where the original copyright information should be kept.
- [x] I'm using the same code standards as the existing code (indentation, spacing, variable naming, ...).
- [ ] I've added tests for my code.
- [x] I have verified whether my change requires changes to the documentation
- [x] My change either requires no documentation change or I've updated the documentation accordingly.
- [x] My branch has been rebased to master, keeping only relevant commits.

PR #225, opened 2019-06-21
Fix compilation with Boost 1.70.0

homespring:
## Summary

I have fixed a build error when using Boost 1.70.0.

## Details

I have added one missing parameter in log_build_info() method overload.

## Motivation and Context

I wanted to use Boost library version 1.70.0 but the compilation failed.

## How Has This Been Tested?

I ran your tests and all of them passed (110/110).

## Types of changes

- [ x] Bug fix (non-breaking change which fixes an issue).
- [ ] New feature (non-breaking change which adds functionality).
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:

- [x ] It is my own work, its copyright is implicitly assigned to the project and no substantial part of it has been copied from other sources (including [Stack Overflow](https://stackoverflow.com/)). In rare occasions this is acceptable, like in CMake modules where the original copyright information should be kept.
- [ x] I'm using the same code standards as the existing code (indentation, spacing, variable naming, ...).
- [ ] I've added tests for my code.
- [ ] I have verified whether my change requires changes to the documentation
- [ x] My change either requires no documentation change or I've updated the documentation accordingly.
- [ ] My branch has been rebased to master, keeping only relevant commits.

PR #230, opened 2019-12-09
Resolves compile error with Boost 1.71.0

dionisioedu:
## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected).

## Checklist:

- [x] I've added tests for my code.


dionisioedu:
There is a new virtual function on the boost::unit_test::unit_test_log_formatter in a new version of boost.

LukasWoodtli:
This is basically a duplicate of #225 and #227. As far as I know the API change is already in Boost 1.70 so the version check is not right in this PR.

PR #231, opened 2019-12-17
Bugfixes

LukasWoodtli:
## Summary

This is a combination of  #228 and #227 

## Details

This combined PR is used to check that both fixes together let the CI builds pass.

## Motivation and Context

Because there are two not 

## How Has This Been Tested?

See the individual PR's.

## Types of changes

## Checklist:

### DIFF
--- a/examples/CalcQt/CMakeLists.txt
+++ b/examples/CalcQt/CMakeLists.txt
@@ -2,6 +2,7 @@ project(CalcQt)
 
 if(TARGET Qt::Core AND TARGET Qt::Gui AND TARGET Qt::Widgets AND TARGET Qt::Test)
     add_library(libcalcqt STATIC src/CalculatorWidget.cpp src/CalculatorWidget.h)
+    target_compile_options(libcalcqt PRIVATE "-Wno-deprecated-declarations")
     set_target_properties(libcalcqt PROPERTIES AUTOMOC ON)
     target_include_directories(libcalcqt INTERFACE src)
     target_link_libraries(libcalcqt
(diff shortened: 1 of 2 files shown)


LukasWoodtli:
@muggenhor, @paoloambrosio, @konserw
Could anyone of you review this. I'm asking you because you are the main contributors...

jermus67:
Hello @LukasWoodtli ,

Maintenance of cucumber-cpp is in the process of being picked up again. The file BoostDriver.cpp has been a popular one to be fixed :). Unfortunately this means that this PR cannot be merged right now. I don't know if you are still active, but if you are, could you update this PR by resolving the conflict for the BoostDriver.cpp? When that is done it can be merged, since the other change is a setting in the QT domain.

There is talk of setting up plugin support for test drivers, so that adding support for frameworks like QT and Google Test, and hence configuring how they are compiled, do not need to be part of cucumber-cpp anymore, but since such a solution is only in the "that might be a good idea" phase any bugfix, also in the test driver support is welcome.

Kind regards,
Jeroen Kouwer

jermus67:
Hallo @LukasWoodtli 

I've taken a leap of faith an merged the other PR (#228, about the QSignalMapper), which saves you the trouble of updating an ancient PR.

Thanks for having provided the fixes.

Kind regards,
Jeroen Kouwer

LukasWoodtli:
Hello @jermus67 
Cool that cucumber-cpp is back on stage!
Unfortunately I am currently not able to participate.
Keep going!

PR #232, opened 2020-01-06
Remove CACHE FORCE arguments from CMAKE_CXX_FLAG on colored terminal output

alexcani:
Remove CACHE FORCE arguments from CMAKE_CXX_FLAG on colored terminal output

## Summary
By using CACHE STRING "" FORCE on 'set' command this updated CMAKE_CXX_FLAG
on any parent directory, thus leaking internal configuration into external
projects that may be using cucumber-cpp through add_directory.

## Details
An example is -Werror -Wall -Wextra flags set on "Generic Compiler Flags".
In a Unix + Ninja + GCC environment those flags would leak into parent
directories

## Types of changes

## Checklist:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -91,7 +91,7 @@ if(UNIX AND (
   # itself is writing to a pipe instead of a terminal. As a result this should give us the best of
   # both worlds: coloured output when we're running in a terminal, plain output for editors, IDEs,
   # etc.
-  set(CMAKE_CXX_FLAGS "-fdiagnostics-color=always ${CMAKE_CXX_FLAGS}" CACHE STRING "" FORCE)
+  set(CMAKE_CXX_FLAGS "-fdiagnostics-color=always ${CMAKE_CXX_FLAGS}")
 endif()
 
 #


jermus67:
@muggenhor, (if you're still reading along and can find some time) can you elaborate on why the ```CACHE STRING "" FORCE``` was part of setting the CMAKE_CXX_FLAGS?

I somehow cannot deduce from the the cmake documentation that this bleeds into the external / calling project, but that may be an implicit property of a cached variable (i.e. always being set at the highest level and not for the current directory only).

aslakhellesoy:
Hi @alexcani,

Thanks for your making your first contribution to Cucumber, and welcome to the Cucumber committers team! You can now push directly to this repo and all other repos under the cucumber organization! 🍾

In return for this generous offer we hope you will:

* ✅ Continue to use branches and pull requests. When someone on the core team approves a pull request (yours or someone else's), you're welcome to merge it yourself.
* 💚 Commit to setting a good example by following and upholding our [code of conduct](https://cucumber.io/conduct) in your interactions with other collaborators and users.
* 💬 Join the [community Slack channel](http://cucumber.io/support#slack) to meet the rest of the team and make yourself at home.
* ℹ️ Don't feel obliged to help, just do what you can if you have the time and the energy.
* 🙋 Ask if you need anything. We're looking for feedback about how to make the project more welcoming, so please tell us!

On behalf of the Cucumber core team,
Aslak Hellesøy
Creator of Cucumber

muggenhor:
> @muggenhor, (if you're still reading along and can find some time) can you elaborate on why the `CACHE STRING "" FORCE` was part of setting the CMAKE_CXX_FLAGS?
> 
> I somehow cannot deduce from the the cmake documentation that this bleeds into the external / calling project, but that may be an implicit property of a cached variable (i.e. always being set at the highest level and not for the current directory only).

Just noticed this message now (ever since our company switched to GitHub I get swamped in GH mails).

Anyway, I'm pretty sure this was a copy paste error on my side because a few other CMake projects that I used at the time injected `CMAKE_*_FLAGS` via the cache, so this was the only way to affect a change there. That wouldn't have applied here, but it affected my approach to thinking about CMake at the time.

PR #234, opened 2020-02-27
Add make option in Readme

RCoeurjoly:
Since sudo cmake doesn't work on my system, I added an alternative


ursfassler:
@RCoeurjoly this PR seems to be closed in error while renaming the main branch. Sorry about that. Please reopen if the changes are still relevant for you.

RCoeurjoly:
Thanks! This PR has very bad code (my path is all over the place), so I think it is best for it to remain closed.
At present I don't use cucumber-cpp, unfortunately, so I am no longer in a hurry to fix this.

PR #235, opened 2020-05-16
Compability with libc: remove -Werror

leonardoInf:
## Summary

Compiling cucumber-cpp using Alpine (libc) produces warnings about Boost using wrong imports.

`/include/sys/poll.h:1:2: error: #warning redirecting incorrect #include <sys/poll.h> to <poll.h> [-Werror=cpp]`

I propose to drop -Werror or otherwise cucumber-cpp will not compile on platforms without glibc.

## Types of changes

- [x]  Bug fix (non-breaking change which fixes an issue).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -65,7 +65,7 @@ option_depr        (VALGRIND_TESTS            CUKE_TESTS_VALGRIND)
 #
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS_INIT} -Werror -Wall -Wextra ${CMAKE_CXX_FLAGS}")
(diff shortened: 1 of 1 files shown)


sappo:
Hi @leonardoInf, would using poll.h instead of sys/poll.h be the bugfix?

Werror prints that message for a reason. Ignoring it by disabling it doesn't seem wise.

leonardoInf:
It´s some internal file from Boost so I do not feel comfortable editing it. I think we should allow warnings to allow builds on different platforms/compilers.

sappo:
This issue was addressed in boost release 1.65. Could you try to compile cucumber-cpp with that or a newer version. 

See boost asio changelog
https://github.com/boostorg/asio/blob/d114cff2d264f38f7eafbdbb73a76707dbb99942/doc/history.qbk#L436

leonardoInf:
Ok, in this PR is not relevant. Thanks for enlightening :D

PR #251, opened 2022-09-30
add Qt on CI system

ursfassler:
## Summary

Execute Qt tests on CI.

## Details

Adding Qt to the CI system.

## Motivation and Context

The Qt tests are not run on the CI system because they are not found. We may not know about a broken state for a long time.

This is a base to add another fix for the latest Qt versions: https://github.com/ursfassler/cucumber-cpp/tree/fix-build-with-qt5.15.2.

## How Has This Been Tested?

With the github action.

## Types of changes

- [x] Bug fix (in the CI system) (non-breaking change which fixes an issue).

## Checklist:

### DIFF
--- a/.github/workflows/ubuntu-latest.yml
+++ b/.github/workflows/ubuntu-latest.yml
@@ -24,6 +24,7 @@ jobs:
     - name: setup environment
       run: |
         sudo apt-get install libboost1.71-all-dev
+        sudo apt-get install qtbase5-dev
   
     - name: Configure CMake
       # Configure CMake in a 'build' subdirectory. `CMAKE_BUILD_TYPE` is only required if you are using a single-configuration generator such as make.


mpkorstanje:
Cheers! Would you mind looking at the appveyor tests too? If they're not duplicates of the current set of tests in GitHub?

ursfassler:
The appveyor tests seem to use Qt, but only run on Windows. All of them [fail](https://ci.appveyor.com/project/cucumberbdd/cucumber-cpp/builds/44931394/job/u34np5c5ykihao2c) during `gem install bundle`.
Is this what you asked for? Should we discuss the whole appveyor topic in a bug report?

mpkorstanje:
That seems reasonable. Though I wouldn't know what to fix about it and wouldn't be able to contribute much. I'm mostly merging everything that looks reasonable to keep momentum going.

PR #254, opened 2022-11-11
use Visual Studio 2022 image for Windows build

ursfassler:
## Types of changes

## Checklist:

- [ ] It is my own work, its copyright is implicitly assigned to the project and no substantial part of it has been copied from other sources (including [Stack Overflow](https://stackoverflow.com/)). In rare occasions this is acceptable, like in CMake modules where the original copyright information should be kept.
- [ ] My branch has been rebased to main, keeping only relevant commits.


.github/workflows/windows.yml
jenisys, line 1: Mmh, I am just wondering why you need a Github action for every C++ package that you need to install as dependency ? Normally, the CMake build should take care of that by using `FetchContent` or a CPM package or … (After just looking at the CMakeLists.txt file in the repository, that does not seem to be the case).

* Would it not be better to extend the CMake build that it fetches its dependencies if needed ?!?
* Installing `boost` normally needs a lot a of time. You can save a lot of time, if you specify which parts of `boost` you are actually using (`components` in CMake terms) except the header-only part of the `boost` library

ursfassler: > Mmh, I am just wondering why you need a Github action for every C++ package that you need to install as dependency ?

Because I don't know better :see_no_evil:

I'll look into the suggested CMake solution. It is not yet done with FetchContent, but if that is the modern way I'd like to go with it as I'd like to modernize the whole project.

I'd be happy to take suggestions or solutions :wink:

jenisys: NOTES:

* `CPM` looks similar to `FetchContent`, but it really simplifies writing `CMakeLists.txt` files
* `CPM` adds no dependency. You bundle it w/ your project and a user can just build the `CMake` project as it is.

SEE ALSO:

* cmake.docs: [FetchContent](https://cmake.org/cmake/help/v3.26/module/FetchContent.html?highlight=fetchcontent)
* https://github.com/cpm-cmake/CPM.cmake
* https://github.com/cogumbreiro/cpm-boost
* `CMake project_options`: https://github.com/aminya/project_options
* CppCon 2017: Mathieu Ropert: [Using Modern CMake Patterns to Enforce Good Modular Design Using Mordern CMake](https://github.com/CppCon/CppCon2017/blob/master/Tutorials/Using%20Modern%20CMake%20Patterns%20to%20Enforce%20a%20Good%20Modular%20Design/Using%20Modern%20CMake%20Patterns%20to%20Enforce%20a%20Good%20Modular%20Design%20-%20Mathieu%20Ropert%20-%20CppCon%202017.pdf)

EXAMPLE:

* Example for CPM and `project_options`: [cpp-best-practices/cmake_template](https://github.com/cpp-best-practices/cmake_template)
* https://github.com/cpp-best-practices/cmake_template/blob/main/Dependencies.cmake

ursfassler:
Closing it because I will focus on modernization instead of make CI work on Windows.

PR #270, opened 2023-12-26
cpp 14 usage

reneFi:
## Summary

googletest has changed to cpp standard 14. To use cucumber-cpp with actual version compile for versions that require CPP 14 is enabled now 
 
## Motivation and Context

use  current versions of googletest framework

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,15 +26,15 @@ set(CUKE_ENABLE_SANITIZER       "OFF" CACHE STRING "Sanitizer to use for checkin
 set_property(CACHE CUKE_ENABLE_SANITIZER PROPERTY STRINGS OFF "address" "thread" "undefined")
 option(CUKE_TESTS_VALGRIND      "Enable tests within Valgrind" OFF)
 set(GMOCK_SRC_DIR "" CACHE STRING "Google Mock framework sources path (otherwise downloaded)")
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -26,15 +26,15 @@ set(CUKE_ENABLE_SANITIZER       "OFF" CACHE STRING "Sanitizer to use for checkin
...
-# according to: https://github.com/google/googletest/tree/release-1.10.0#18x-release
-# the 1.8.x is the last release that works with pre-C++11 compilers.
-if(GMOCK_VER VERSION_GREATER_EQUAL "1.9.0")
+# according to: https://github.com/google/googletest/releases/tag/v1.13.0
+# the 1.12.x is the last release that works with pre-C++14 compilers.
+if(GMOCK_VER VERSION_GREATER_EQUAL "1.14.0")
     if(NOT DEFINED CMAKE_CXX_STANDARD)
-        set(CMAKE_CXX_STANDARD 11)
-    elseif(CMAKE_CXX_STANDARD LESS 11)
-        message(FATAL_ERROR "C++11 (above) is required by googletest version: ${GMOCK_VER}")
```
ursfassler, line 37: With the recent changes we need C++17 for cucumber itself. We shouldn't need this section anymore (it's also the one with the merge conflict).

ursfassler (changes requested):
Thanks for the changes. Please also adjust the Readme and see the comments below.

PR #272, opened 2023-12-27
remove broken E2E tests

ursfassler:
## Summary

Remove broken E2E tests.

## Motivation and Context

The E2E/feature-tests depend on the outdated Cucumber Technology Compatibility Kit.

## Types of changes

## Checklist:


luke-hill:
FYI the CCK has been updated a lot recently. It's likely that to get things working / passing you'll need a "set" of updates to all the ruby based technologies at the same time.

I know you're still firefighting a lot here, so let me know if I can be of any help and I'll try where possible to help.

The `wire` gem "should" now be almost fully re-conformant and using a semi up to date version of cucumber ruby

ursfassler:
The E2E test used the archived [Cucumber Technology Compatibility Kit (TCK)](https://github.com/cucumber-attic/cucumber-tck). The [CCK (Cucumber Compatibility Kit)](https://github.com/cucumber/compatibility-kit) does verify a bit different things and is not used by cucumber-cpp (please correct me if I am mixing up something, it is a bit confusing).

I'd like to integrate CCK, but it seems to only make sense once we support [Cucumber Expressions](https://github.com/cucumber/cucumber-expressions#readme) (that would be awesome, because the regexes are bugging me the most when I'm using cucumber-cpp in a project).

luke-hill:
Sorry you are right. I missed something. Ignore my comment, I was wrong. You are indeed right about it all.

Next step imo is to establish some semi-conformance to using up to date ruby techs. I can either create an issue or let you attack it head on. Either is fine by me.

ursfassler:
I am not sure what you are talking about :see_no_evil: => would be good to have an issue to explain it a bit.

PR #285, opened 2024-01-05
wip: check CI run with cucumber 9 and cucumber-wire 7

ursfassler:
## Types of changes

## Checklist:

- [ ] It is my own work, its copyright is implicitly assigned to the project and no substantial part of it has been copied from other sources (including [Stack Overflow](https://stackoverflow.com/)). In rare occasions this is acceptable, like in CMake modules where the original copyright information should be kept.
- [ ] My branch has been rebased to main, keeping only relevant commits.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,7 +1,7 @@
 source 'https://rubygems.org'
 
 group :test do
-  gem 'cucumber', "=7.1.0"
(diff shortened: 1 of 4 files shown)


ursfassler:
Have to extend the test

PR #293, opened 2024-04-08
adding windows test driver fix

kreuzberger:
## Types of changes

## Checklist:

### DIFF
--- a/src/drivers/QtTestDriver.cpp
+++ b/src/drivers/QtTestDriver.cpp
@@ -8,25 +8,32 @@ namespace cucumber {
 namespace internal {
 
 const InvokeResult QtTestStep::invokeStepBody() {
-    QTemporaryFile file;
-    if (!file.open()) {
-        return InvokeResult::failure("Unable to open temporary file needed for this test");
+    QString file_name;
+    {
+        QTemporaryFile file;
+        if (!file.open()) {
+            return InvokeResult::failure("Unable to open temporary file needed for this test");
+        }
+        file.close();
     }
-    file.close();
 
-    QtTestObject testObject(this);
+    file_name += ".txt";
+
+    QtTestObject testObject( this );
     int returnValue = QTest::qExec(
         &testObject,
         QStringList() << "test"
-                      << "-o" << file.fileName()
+                      << "-o" << file_name
     );
-    if (returnValue == 0) {
-        return InvokeResult::success();
-    } else {
-        file.open();
-        QTextStream ts(&file);
-        return InvokeResult::failure(ts.readAll().toLocal8Bit());
+    auto ok = ( returnValue == 0 ) ? true : false;
+    std::string error_text;
+    QFile file( file_name );
+    if(!ok) {
+      file.open( QIODevice::ReadOnly );
+      error_text = QString::fromLocal8Bit( file.readAll() ).toStdString();
     }
+    //file.remove();
+    return ok ?  InvokeResult::success() :  InvokeResult::failure( error_text );
 }
 
 }


src/drivers/QtTestDriver.cpp
```diff
@@ -8,25 +8,32 @@ namespace cucumber {
...
+        QTemporaryFile file;
+        if (!file.open()) {
+            return InvokeResult::failure("Unable to open temporary file needed for this test");
+        }
+        file.close();
     }
-    file.close();
 
-    QtTestObject testObject(this);
+    file_name += ".txt";
```
ursfassler, line 20: The file name is always ".txt" as it is not derived from the temporary file (I'd expect this was the intention).

src/drivers/QtTestDriver.cpp
```diff
@@ -8,25 +8,32 @@ namespace cucumber {
...
-    QTemporaryFile file;
-    if (!file.open()) {
-        return InvokeResult::failure("Unable to open temporary file needed for this test");
+    QString file_name;
+    {
+        QTemporaryFile file;
+        if (!file.open()) {
+            return InvokeResult::failure("Unable to open temporary file needed for this test");
+        }
+        file.close();
```
ursfassler, line 17: Closing the file here can lead to bugs. In the sequence between here and when the file is opened again to read, another process could create a file with the same name. Is it possible to open the file for reading while another process is writing to it?

src/drivers/QtTestDriver.cpp
```diff
@@ -8,25 +8,32 @@ namespace cucumber {
...
-                      << "-o" << file.fileName()
+                      << "-o" << file_name
     );
-    if (returnValue == 0) {
-        return InvokeResult::success();
-    } else {
-        file.open();
-        QTextStream ts(&file);
-        return InvokeResult::failure(ts.readAll().toLocal8Bit());
+    auto ok = ( returnValue == 0 ) ? true : false;
```
ursfassler, line 28: Make it const and remove true and false:
```C++
const auto ok = returnValue == 0;
```

src/drivers/QtTestDriver.cpp
```diff
@@ -8,25 +8,32 @@ namespace cucumber {
...
+    std::string error_text;
+    QFile file( file_name );
+    if(!ok) {
+      file.open( QIODevice::ReadOnly );
+      error_text = QString::fromLocal8Bit( file.readAll() ).toStdString();
     }
+    //file.remove();
+    return ok ?  InvokeResult::success() :  InvokeResult::failure( error_text );
```
ursfassler, line 36: Imho the readability of the old structure was better. Maybe something like:

```C++
  if(returnValue == 0) {
    return InvokeResult::success();
  }

  QFile file( file_name );
  file.open( QIODevice::ReadOnly );
  const auto error_text = QString::fromLocal8Bit( file.readAll() ).toStdString();

  return InvokeResult::failure( error_text );
}
```

src/drivers/QtTestDriver.cpp
```diff
@@ -8,25 +8,32 @@ namespace cucumber {
...
-        QTextStream ts(&file);
-        return InvokeResult::failure(ts.readAll().toLocal8Bit());
+    auto ok = ( returnValue == 0 ) ? true : false;
+    std::string error_text;
+    QFile file( file_name );
+    if(!ok) {
+      file.open( QIODevice::ReadOnly );
+      error_text = QString::fromLocal8Bit( file.readAll() ).toStdString();
     }
+    //file.remove();
```
ursfassler, line 35: Don't check in commented code.

ursfassler (changes requested):
Thanks for the contribution!

I see the problem and how you solve it. Please fix and adjust your changes according my input (or discuss it when you see it different :slightly_smiling_face: ).

Please also add the test in the CI. This helps me to see the problem and protects us from similar problems in the future.

src/drivers/QtTestDriver.cpp
kreuzberger: right, should be file_name = file.fileName() + ".txt"

src/drivers/QtTestDriver.cpp
kreuzberger: This is the major problem. The file MUST go out of scope (this means also implicitly closed()) for the test to have write access on windows.
Cause the filename as the process name as file name part, another process should not have the same name.
Considering multiple parallel tests of the same executable is an issue, could be fixed by adding the pid in the file name template.

src/drivers/QtTestDriver.cpp
kreuzberger: See above :smile:

src/drivers/QtTestDriver.cpp
kreuzberger: depends on the comment :smile: 
Currently the files (temporay) remain existing after test finished. This is no problem on linux where the /tmp/ is cleaned. on Windows those file fragments will remain. 
Should they be removed after test? 
[yes] -> no early exit or remove in early exit code
[no] -> then your suggestion is accepted.

kreuzberger:
Due to the checks: Could you provide a clang-format file (.clang-format) or what should be used for clang-format as default style?

ursfassler:
> Due to the checks: Could you provide a clang-format file (.clang-format) or what should be used for clang-format as default style? Assume clang-format without any further options?

The format file is checked in: https://github.com/cucumber/cucumber-cpp/blob/main/.clang-format
Clang-format may produce slightly different out depending on the version. It usually is no problem.

src/drivers/QtTestDriver.cpp
ursfassler: Ok, I see the problems. According to the documentation, it does indeed look like there is no easy solution.

As it gets more tricky a helper class may be needed. Something like (I haven't compiled nor test the code, needs also one comment to explain why we do it in such a complicated manner):
```C++

class TemporaryFileWrapper {
public:
    static std::optional<TemporaryFileWrapper> create()
    {
        QTemporaryFile tempFile{};
        tempFile.setAutoRemove(false);

        if(!tempFile.open()) {
            return {};
        }

        return {tempFile.fileName() + ".txt"};
    }

    TemporaryFileWrapper(QString filename): filename{filename} {
    }

    ~TemporaryFileWrapper() {
        QFile::remove(filename);
    }

    QString name() const {
        return filename;
    }

    QString read() const {
        QFile file{ filename };
        if (!file.open(QIODevice::ReadOnly | QIODevice::Text))
            return QString();
        QTextStream in(&file);
        return in.readAll();
    }

private:
    QString filename;
};


const InvokeResult QtTestStep::invokeStepBody() {
    const auto file = TemporaryFileWrapper::create();
    if (!file) {
        return InvokeResult::failure("Unable to open temporary file needed for this test");
    }
    
    QtTestObject testObject(this);
    int returnValue = QTest::qExec(
        &testObject,
        QStringList() << "test"
                      << "-o" << file->name()
    );
    
    if (returnValue == 0) {
        return InvokeResult::success();
    } else {
        return InvokeResult::failure(file->read().toLocal8Bit());
    }
}
```

kreuzberger: A solution could be to use the standard function tmpnam. So no file is generated, it is just the name generated.
The template should also be application name and pid. But this always gives me a warning cause considered insecure) which should be ignored.

kreuzberger: Saw your suggestion to late :smile: Would give it a try tomorrow.

kreuzberger: Tried to integrate the wrapper without std::optional.
Open Issues: if Autoremove is disabled, the temporary file is NOT removed with QFile::remove. Still existing.
The return value is true (So file should be removed) but is still "there" in the temp folder.
Autoremove in "normal" behaviour fixes this. Cause we do want only the filename, this would be ok for me.

kreuzberger:
Hi! Would it be possible to let the checks run on commit, so i can see open issues and fix them?

src/drivers/QtTestDriver.cpp
```diff
@@ -4,28 +4,81 @@
...
+class TemporaryFileWrapper {
+public:
+    static TemporaryFileWrapper create() {
+        QTemporaryFile tempFile(QString("%1/%2_%3")
+                                    .arg(
+                                        QDir::tempPath(),
+                                        qApp->applicationName().isEmpty() ? "qt_temp"
+                                                                          : qApp->applicationName()
+                                    )
+                                    .arg(qApp->applicationPid()));
```
ursfassler, line 24: According to https://doc.qt.io/qt-6/qtemporaryfile.html#QTemporaryFile the default constructor creates unique filename. Doesn't this work on Windows the same way?

tests/CMakeLists.txt
```diff
@@ -11,6 +11,37 @@ target_include_directories(utils INTERFACE
...
+    if( NOT "${CUKE_ENV_BINPATH}" STREQUAL "")
+        if(WIN32)
+            string(REPLACE ";" "\;" binpath "${CUKE_ENV_BINPATH}")
+        endif()
+        list(APPEND environment "PATH=path_list_prepend:$<SHELL_PATH:${binpath}>")
+    endif()
+    
+    set(${environment} ${${environment}} PARENT_SCOPE)
+endfunction()
+
```
ursfassler: Why is this needed? is it to run the tests under windows with dynamic linked cucumber libraries?
It bothers me that so much cmake code is needed, I fear it is brittle and gets difficult to maintain. I'd like to see a more straight forward solution (i.e. more of a standard solution, less cmake code) if possible.

ursfassler (commented):
I looked through the changes shortly to see if I can run the tests, i.e. that no bad things are in there :smile:

I haven't really understood the problem with Qt, Windows and tests, hence I have troubles to understand if the patch is good or not. Sorry for that, a bit short on time on my end.

ursfassler:
> On Windows the QtTestDriver failes sometimes. Reason is that in one test the Text Error messages are compared.
Due to the nature of the default QtTest output the diff test fails cause the Test execution time differs.

:astonished: That would be a bad tests, timings shouldn't be compared in the tests. What test is it?

ursfassler:
> Fixed run-linux.sh issue to use Qt6 and disable Qt5 if both are found

Idea is that only the Qt version you specified is searched for.

src/drivers/QtTestDriver.cpp
kreuzberger: This code was to ensure that it behaves equal between qt5 and qt6 and your concerns about running tests in parallel. 
QTemporaryFile between Qt5 and Qt6 seems to differ in default temp file name

tests/CMakeLists.txt
kreuzberger: running the tests on windows would not be possible if qt is not in PATH. I do not know or found any code how this is set in your workflow file.
This is what the cmake code is for.
I can reduce it to RUNPATH handling only. But if not there i have a strong feeling that the tests would not run on windows.

kreuzberger:
> > On Windows the QtTestDriver failes sometimes. Reason is that in one test the Text Error messages are compared.
> > Due to the nature of the default QtTest output the diff test fails cause the Test execution time differs.
> 
> 😲 That would be a bad tests, timings shouldn't be compared in the tests. What test is it?

QtTestDirverTest

ursfassler:
@kreuzberger please see my PR #296 to fix the flaky test.

I also investigated if it is possible to capture the test log without needing the file. Unfortunately it seems not to be possible (without too much hacking). Meaning that this PR is quite relevant.

kreuzberger:
> @kreuzberger please see my PR #296 to fix the flaky test.
> 
> I also investigated if it is possible to capture the test log without needing the file. Unfortunately it seems not to be possible (without too much hacking). Meaning that this PR is quite relevant.

Comment your PR#296.

ursfassler:
@kreuzberger unfortunetaly the Windows build has a problem with gtest.

@cwellm did you also had this problem? Could you also help out with reviewing the windows code?

tests/CMakeLists.txt
```diff
@@ -11,6 +11,30 @@ target_include_directories(utils INTERFACE
...
+    if( NOT "${CUKE_ENV_RUNPATH}" STREQUAL "")
+        if(WIN32)
+            string(REPLACE ";" "\;" CUKE_ENV_RUNPATH "${CUKE_ENV_RUNPATH}")
+        endif()
+        set(RUNPATH "$<IF:$<PLATFORM_ID:Windows>,PATH,LD_LIBRARY_PATH>")
+        list(APPEND environment "$<IF:$<PLATFORM_ID:Windows>,PATH,LD_LIBRARY_PATH>=path_list_prepend:$<SHELL_PATH:${CUKE_ENV_RUNPATH}>")
+    endif()
+    set(${environment} ${${environment}} PARENT_SCOPE)
+endfunction()
+
```
ursfassler, line 37: This is a lot of code to just support Windows. Is it possible to move this to run-windows.ps1 (if it makes sense).

@cwellm any ideas from your side?

ursfassler (commented):
@kreuzberger  I think we get closer on your fix but still have troubles with the CI. I'd suggest that you split your changes to a small fix for the temporary file on Windows and we add the tests in a separate PR. This way you have the changes that you need integrated.

If you agree, I'd suggest that you open a second PR with the Windows-tests and only keep the functional changes here. Please also cleanup the history in order to only have relevant commits in it.

src/drivers/QtTestDriver.cpp
```diff
@@ -4,28 +4,81 @@
...
+
+        if (!tempFile.open()) {
+            return {};
+        }
+
+        return {tempFile.fileName() + ".txt"};
```
ursfassler, line 30: We may have a misunderstand here. According to the QTemporaryFile documentation for Qt5 and Qt6, the default constructor (without arguments) is guaranteed to create a unique filename. I think we should use this because creating unique filenames can be difficult, and I guess Qt did it good. It is ok for me when the filename is different between runs and Qt versions, as long as the content is deterministic (what it is by now).

My suggestion is:
```C++
    static TemporaryFileWrapper create() {
        QTemporaryFile tempFile();

        if (!tempFile.open()) {
            return {};
        }

        tempFile.setAutoRemove(false); // this is important if we remove the .txt extension! Otherwise the temp file is deleted at the end of this function. Remove this comment before committing :)

        return tempFile.fileName();
    }
```

CMakeLists.txt
```diff
@@ -98,7 +98,7 @@ endif()
 #
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS_INIT} -Werror -Wall -Wextra -Wsuggest-override ${CMAKE_CXX_FLAGS}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS_INIT} -Wall -Wextra -Wsuggest-override ${CMAKE_CXX_FLAGS}")
```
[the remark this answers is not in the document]
ursfassler: I removed the -Werror with #301.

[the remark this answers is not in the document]
kreuzberger: Reason to remove the flag was NOT the code from cucumber-cpp, the error came via asio.

PR #296, opened 2024-04-29
fix flaky QtTestDriver test

ursfassler:
## Summary

The test output contained timing information, leading to a flaky test. Fixed by using the TAP output, that doesn't contain timing information.

## Details

See https://doc.qt.io/qt-6/qtest-overview.html#logging-options for more information.

## Motivation and Context

Solves the issue from this PR: https://github.com/cucumber/cucumber-cpp/pull/293#issuecomment-2049850379

## How Has This Been Tested?

The fix itself is only manually verified. The code has automated tests, hence pretty sure to not have added a regression.

**Note**: the log output changed, but it should only affect developers reading the log.

## Types of changes

## Checklist:

### DIFF
--- a/src/drivers/QtTestDriver.cpp
+++ b/src/drivers/QtTestDriver.cpp
@@ -14,12 +14,10 @@ const InvokeResult QtTestStep::invokeStepBody() {
     }
     file.close();
 
-    QtTestObject testObject(this);
-    int returnValue = QTest::qExec(
-        &testObject,
-        QStringList() << "test"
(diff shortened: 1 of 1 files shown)


ursfassler:
@kreuzberger please have a look. Would be happy to get feedback if this fixes the flaky tests you encountered.

kreuzberger:
> @kreuzberger please have a look. Would be happy to get feedback if this fixes the flaky tests you encountered.

as the tap format does not contain timing informations, yes this would fix it.
:+1:

Issue #295, opened 2024-04-09
Compilation and Library Use


kreuzberger:
### Compilation
- Compilation without `.git ` (e.g. from downloaded Source Archive) leads to error cause `CUKE_VERSION` could not be determined
- Use Library in a Shared Library requires compilation with `-fPIC` on Unix
- Compilation error due to `-Werror`:
``` 
include/asio/impl/error.ipp:34:15: error: ‘virtual const char* asio::error::detail::netdb_category::name() const’ can be marked override [-Werror=suggest-override]
```

### ✨ Do you have a proposal for making it better?

* If compiled from source archive the used **CUKE_VERSION** must be SET via **Configure** OR the source archive should contain a version file from which CMake could read the **CUKE_VERSION**
* Compile with flag -fPIC in Unix cases
* Remove -Werror, No warnings goal should be checked via github actions on merge / build

### 📚 Any additional context?

Below suggestion to remove the issues without greater changes

```
diff --git a/CMakeLists.txt b/CMakeLists.txt
index cc98353..7b92ff4 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -98,7 +98,7 @@ endif()
 #
 
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang|GNU")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS_INIT} -Werror -Wall -Wextra -Wsuggest-override ${CMAKE_CXX_FLAGS}")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS_INIT} -Wall -Wextra -Wsuggest-override ${CMAKE_CXX_FLAGS}")
     # TODO: A better fix should handle ld's --as-needed flag
     if(UNIX AND NOT APPLE)
         set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Xlinker '--no-as-needed'")
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index 466c37c..cda2006 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -127,7 +127,18 @@ foreach(TARGET
     endif(MINGW)
 endforeach()
 
-git_get_version(CUKE_VERSION)
+if(UNIX)
+    foreach(TARGET
+        cucumber-cpp-internal
+        cucumber-cpp-nomain
+    )
[...]

PR #300, opened 2024-05-09
allow compilation without git or when git fails

ursfassler:
## Summary

Allow compilation without git or when git fails.

## Details

This happens when the code is downloaded as source archive. This does not contain the git repository, hence the git version script fails. 

Now the version 0.0.0+YEAR.MONTH.DAY is used when we fail to get the version from git.

## Motivation and Context

One problem mentioned in #295.

## How Has This Been Tested?

Manually by removing the .git directory and adding it back again.

## Types of changes

## Checklist:

### DIFF
--- a/cmake/modules/GitVersion.cmake
+++ b/cmake/modules/GitVersion.cmake
@@ -1,8 +1,12 @@
 function(git_get_version VERSION_VARIABLE)
+    string(TIMESTAMP TODAY "%Y.%m.%d")
+    set(${VERSION_VARIABLE} "0.0.0+${TODAY}" PARENT_SCOPE)
+
     find_program(GIT_EXECUTABLE git)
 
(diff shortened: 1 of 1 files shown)


ursfassler:
@kreuzberger please let me know if this works for the compilation problem mentioned in #295.

kreuzberger:
Technically this solves my problem using the downloaded zip/tar archive for the tag.

But is this the real intention of your software deployment? As long as i am using Tags of your library/executable, i would expect to see the correct version number of the software/library. So maybe the software version should be hardcoded and not be 0.0.0 between the tags. It would be okay to add the timestamp/date information (0.7.0+ddmmyy) so that is clear this is not an "offical" git tag build.

Another way could be a "released" version of your code, where the version comes from a file generated during the release build.

Up to you :+1:

PR #309, opened 2025-06-13
fix: verify boost url and print content in the directory

shirzartenwer:
## Summary
The original boost download URL was not working anymore. Updated to a new one.

## Motivation and Context

Bug fix.

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue).

## Checklist:

- [X] It is my own work, its copyright is implicitly assigned to the project and no substantial part of it has been copied from other sources (including [Stack Overflow](https://stackoverflow.com/)). In rare occasions this is acceptable, like in CMake modules where the original copyright information should be kept.
- [X] I'm using the same code standards as the existing code (indentation, spacing, variable naming, ...).
- [X] My change either requires no documentation change or I've updated the documentation accordingly.
- [X] My branch has been rebased to main, keeping only relevant commits.

### DIFF
--- a/.github/workflows/windows-build.yml
+++ b/.github/workflows/windows-build.yml
@@ -59,9 +59,13 @@ jobs:
       if: steps.cache-boost-deps.outputs.cache-hit != 'true'
       run: |
         $boost_version_str = ${Env:BOOST_VERSION}.Replace(".","_")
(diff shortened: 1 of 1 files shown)


ursfassler:
Thank you @shirzartenwer , very good to have that working again.

